PR #5, opened 2021-03-20
Better attestation exception handling

voloshyn:
### Motivation
Attested Clients encrypt/decrypt functions use an internal counter for each communication message, so if a network error prevents an encrypted message from being delivered, the local and remote internal counters will go out of sync and the attestation state needs to be reset.

### In this PR
* Clean up Attestation and Network Exceptions handling
* Re-attest and retry GRPC calls automatically for internal and permission denied errors

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
@@ -144,7 +144,6 @@ public class OwnedTxOut implements Serializable {
 
     @NonNull
     UnsignedLong getTxOutGlobalIndex() {
-        Logger.i(TAG, "Getting txOutGlobalIndex", null, txOutGlobalIndex);
         return txOutGlobalIndex;
     }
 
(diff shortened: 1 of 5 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
```diff
@@ -144,15 +144,13 @@ synchronized void setSpent(
...
     UnsignedLong getTxOutGlobalIndex() {
-        Logger.i(TAG, "Getting txOutGlobalIndex", null, txOutGlobalIndex);
         return txOutGlobalIndex;
     }
 
     int getKeyImageHashCode() {
         if (keyImageHash == 0) {
             keyImageHash = Arrays.hashCode(keyImage);
         }
-        Logger.d(TAG, "Getting KeyImageHashCode", null, keyImageHash);
```
voloshyn, line 155: these getter logs are very spammy

sugargoat (approved):

PR #18, opened 2021-04-02
Remove reportUri normalization

voloshyn:
### Motivation

Remove `reportUri` normalization to use the exact input parameters for AccountKey and PublicAddress creation. This is needed to make sure serialization/deserialization does not change the serialized payload bytes.

Another consequence of this change is that equivalent `PublicAddresses` with implicit and explicit default ports will no longer be equal and will generate different b58 representations.

### DIFF
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
@@ -2,6 +2,8 @@
 
 package com.mobilecoin.lib;
 
+import android.net.Uri;
+
 import androidx.test.platform.app.InstrumentationRegistry;
 
 import com.mobilecoin.lib.exceptions.InvalidUriException;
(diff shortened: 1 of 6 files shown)


android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
```diff
@@ -99,10 +101,88 @@
...
+        RistrettoPrivate key2_copy = RistrettoPrivate.fromBytes(key2.getKeyBytes());
+
+        AccountKey first = new AccountKey(
+                key1,
+                key2,
+                fogUri,
+                fogConfig.getFogReportId(),
+                fogConfig.getFogAuthoritySpki()
+        );
+        try {
```
samdealy, line 123: What's the advantage of writing it this way as opposed to "Assert.assertNotNull(first)"?

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
```diff
@@ -99,10 +101,88 @@
...
+        } catch (Exception e) {
+            Assert.fail(e.toString());
+        }
+        AccountKey second = new AccountKey(key1_copy,
+                key2_copy,
+                fogUri,
+                fogConfig.getFogReportId(),
+                fogConfig.getFogAuthoritySpki()
+        );
+        if (!first.equals(second)) {
```
samdealy, line 136: What's the advantage of writing it this way over "Assert.assertEquals("Public addresses with copied keys must be qual", first, second)"?

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
```diff
@@ -114,6 +194,29 @@ public void test_serialize() throws SerializationException, InvalidUriException
...
+        Uri fogUriWithPort = Uri.parse("fog://some-test-uri:443");
+        Uri[] fogUrisToTest = new Uri[] {fogUri, fogUriWithPort};
+
+        for (Uri uri : fogUrisToTest) {
+            AccountKey accountKey = AccountKey.createNew(
+                    uri,
+                    fogConfig.getFogReportId(),
+                    fogConfig.getFogAuthoritySpki()
+            );
+            byte[] serialized = accountKey.toByteArray();
```
samdealy, line 209: Consider renaming "serialized" to "serializedAccountKey."

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
```diff
@@ -114,6 +194,29 @@ public void test_serialize() throws SerializationException, InvalidUriException
...
+            AccountKey restored = AccountKey.fromBytes(serialized);
```
samdealy, line 210: Consider renaming "restored" to "restoredAccountKey."

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
voloshyn: `first` is valid already since it's returned from the NonNull API, but I'm testing `equals` method to be able to compare to `null` without any exceptions

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
voloshyn: it's equivalent

sugargoat (approved):

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
```diff
@@ -99,10 +101,88 @@
         return accountTestData;
     }
 
+    @Test
+    public void test_compare() throws InvalidUriException, SerializationException {
```
samdealy, line 105: I think it might be helpful to break this test into multiple unit tests. Right now it's testing the "equals" method in five scenarios. I think it's best to keep unit tests as small as possible, and to only invoke the method under test once. This helps the reader understand what the test is about.

That said, consider splitting this test up into multiple unit tests. One for each time equals or equivalent is called. WDYT?

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
samdealy: In that case, consider using "Assert.assertEquals" because it reduces complexity by removing an if statement.

android-sdk/src/main/java/com/mobilecoin/lib/AccountKey.java
```diff
@@ -450,13 +446,38 @@ public String getFogReportId() {
...
+        return Objects.equals(fogReportId, that.fogReportId) &&
+                getNormalizedFogReportUri().equals(that.getNormalizedFogReportUri()) &&
+                Arrays.equals(fogAuthoritySpki, that.fogAuthoritySpki) &&
+                viewKey.equals(that.viewKey) &&
+                spendKey.equals(that.spendKey) &&
+                subAddressViewKey.equals(that.subAddressViewKey) &&
+                subAddressSpendKey.equals(that.subAddressSpendKey);
+    }
+
     @Override
```
samdealy, line 474: This isn't a blocking comment, but somthing to note for the future. It might be worth exploring using [AutoValue](https://github.com/google/auto/tree/master/value) for this class. AutoValue provides `equals` and `hashCode` for free, and we wouldn't need to worry about unit testing that these methods work. This would help reduce boilerplate and number of unit tests. 

Can discuss more offline and also in relation to my unit testing project.

voloshyn: sounds good, let's discuss this offline

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
voloshyn: I think that's a good idea. Almost all of the tests have multiple use-cases in them, there are some pros and cons, let's discuss this offline and add a refactoring ticket for what we decide.

samdealy (approved):

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
voloshyn: done

PR #28, opened 2021-05-12
Create mobilecoin client interface

samdealy:
### Motivation

Currently, there's no easy way for host apps to unit test the MobileCoinClient because it can't be mocked (marked as final) and creating an instance requires complicated configuration. 

### In this PR
This PR extracts MobileCoinClient public methods into an interface. This enables host apps to implement test objects that implement the itnerface in their unit tests. 

### Future Work
Continue extracting interfaces. Also refactor how MobileCoinClientImpl exposes the different clients (fogClient, viewClient, etc.) to make us not have to cast MobileCoinClient types into MobileCoinClientImpl.

[Android-88](https://mobilecoin.atlassian.net/browse/ANDROID-88)

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClientImpl.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClientImpl.java
@@ -1,0 +1,694 @@
+// Copyright (c) 2020-2021 MobileCoin. All rights reserved.
+
+package com.mobilecoin.lib;
+
+import android.net.Uri;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.mobilecoin.api.MobileCoinAPI;
+import com.mobilecoin.lib.exceptions.AttestationException;
+import com.mobilecoin.lib.exceptions.FeeRejectedException;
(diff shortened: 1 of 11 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClientImpl.java
```diff
@@ -0,0 +1,694 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+
+import consensus_common.ConsensusCommon;
+import fog_ledger.Ledger;
+
+/** The default implementation for the {@link MobileCoinClient}. */
+public final class MobileCoinClientImpl implements MobileCoinClient {
```
voloshyn, line 48:
```suggestion
public final class MobileCoinClient implements MobileCoinClientInterface {
```

voloshyn (changes requested):

samdealy:
Opted to make multiple interfaces rather than just one. https://github.com/mobilecoinofficial/android-sdk/pull/29

PR #29, opened 2021-05-12
Create new interfaces for the MobileCoinClient functionality

samdealy:
Motivation
Currently, there's no easy way for host apps to unit test the MobileCoinClient functionality because it can't be mocked (marked as final) and creating an instance requires complicated configuration.

In this PR
This PR extracts MobileCoinClient public methods into three interfaces. This enables host apps to implement test objects that implement the required interfaces in their unit tests.

Future Work
Continue extracting interfaces. Create similarly-named interfaces for the AccountSnapshot class, which basically does everything MobileCoinClient does, but without making network requests. 

Android-88

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinAccountClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinAccountClient.java
@@ -35,7 +35,7 @@ public interface MobileCoinAccountClient {
    */
   @Nullable
   AccountSnapshot getAccountSnapshot(UnsignedLong blockIndex) throws NetworkException,
-      InvalidFogResponse, AttestationException
+      InvalidFogResponse, AttestationException;
 
   /**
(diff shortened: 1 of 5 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinAccountClient.java
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Returns whether the defragmentation is required on the active account in order to send the
+   * specified amount
+   */
+  boolean requiresDefragmentation(@NonNull BigInteger amountToSend)
+      throws NetworkException, InvalidFogResponse, AttestationException,
+      InsufficientFundsException;
+
+  /**
+   * Defragments the user's account.
```
samdealy, line 55: This is a Github text parsing error. The comments are syntactically correct and Android Studio displays it without any errors.

voloshyn (approved):

PR #32, opened 2021-06-09
Update android bindings to v1.1.0

voloshyn:
### In this PR
* Update android bindings to v1.1.0


jcape (commented):
armabi-v7a isn't needed?

voloshyn:
> armabi-v7a isn't needed?

yes, it is there as well

sugargoat (approved):

PR #34, opened 2021-06-28
Construct Amount without CompressedCommitment.

samdealy:
### Motivation
Fog no longer stores a compressed commitment in a TxOutRecord, and the Android SDK's OwnedTxOut constructor uses this compressed commitment to construct an Amount object. As such, we need a new way to create an Amount object without a compressed commitment.

### In this PR
This PR creates a new Amount constructor that doesn't use a compressed commitment. This constructor calls into the native "android-bindings" libs, so this PR contains the necessary Java native interface methods and contains the compiled "android-bindings" libs. 

Fixes [ANDROID-87](https://mobilecoin.atlassian.net/browse/ANDROID-87?atlOrigin=eyJpIjoiMTMyYmI5ZjhhODFhNDEwMmEzYjg4NmVmNTE1NjA4NWIiLCJwIjoiaiJ9).

### Future Work
Update the Fog repo's android-bindings directory, which contains the Rust code that compiles into the .so libs included in this PR.

On the Fog side, [PR 99](https://github.com/mobilecoinfoundation/fog/pull/99) adds the Rust methods that implement that native method interfaces.


voloshyn (commented):
Please link android-bindings PR for reference

samdealy:
> Please link android-bindings PR for reference

Done.

android-sdk/src/main/java/com/mobilecoin/lib/Amount.java
```diff
@@ -43,6 +43,28 @@
         }
     }
 
+    /**
+     * Constructs native Amount object from the txOutSharedSecret and masked value.
+     *
+     * @param txOutSharedSecret  A {@link RistrettoPrivate} representing the shared secret.
```
voloshyn, line 49: this has to be RistrettoPublic
```suggestion
     * @param txOutSharedSecret  A {@link RistrettoPublic} representing the shared secret.
```

voloshyn (approved):
LGTM

android-sdk/src/main/java/com/mobilecoin/lib/Amount.java
samdealy: Done.

PR #35, opened 2021-07-09
Allow plain text communication for insecure protocols

voloshyn:
### Motivation

Some types of testing require non-TLS communication.

### In this PR
* Enable plain text for `insecure-service_name://` protocols
* Use generics for specific service Uris

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/AttestedClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/AttestedClient.java
@@ -15,6 +15,7 @@ import com.mobilecoin.lib.log.Logger;
 import com.mobilecoin.lib.network.services.http.clients.RestClient;
 import com.mobilecoin.lib.network.services.transport.GRPCTransport;
 import com.mobilecoin.lib.network.services.transport.Transport;
+import com.mobilecoin.lib.network.uri.MobileCoinUri;
 
 import attest.Attest;
 import io.grpc.ManagedChannel;
@@ -23,7 +24,7 @@ import io.grpc.ManagedChannel;
  * Base class for attested communication with View/Ledger/Consensus servers
  */
 
-abstract class AttestedClient extends AnyClient {
+abstract class AttestedClient<MobUri extends MobileCoinUri> extends AnyClient<MobUri> {
     private final static String TAG = AttestedClient.class.getName();
     // How long to wait for the managed connection to gracefully shutdown in milliseconds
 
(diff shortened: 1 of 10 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/AttestedClient.java
```diff
@@ -23,7 +24,7 @@
  * Base class for attested communication with View/Ledger/Consensus servers
  */
 
-abstract class AttestedClient extends AnyClient {
+abstract class AttestedClient<MobUri extends MobileCoinUri> extends AnyClient<MobUri> {
     private final static String TAG = AttestedClient.class.getName();
```
samdealy, line 28: I've only seen generic parameter, such as "T" or "U" be used inside of the <>. E.g. AnyClient<T extends MobileCoinUri>. 

Why does it need to be this way, and what does it express? Is there a way to rewrite it such that it uses a generic param such as "T"?

samdealy (commented):
Left a comment regarding the <MobUri extends MobilcCoinUri> usage.

android-sdk/src/main/java/com/mobilecoin/lib/AttestedClient.java
voloshyn: Yes, it can be `T` or anything else for that matter, but it is more informative and easier to read `MobUri uri;` vs `T uri;`
`K, N, T, V, U, S, etc` are from an old Oracle convention, another convention is to use prefix/suffix with a normal type name, e.g. `MobUriT` or `TMobUri`, but it's also old and the reasons for it don't apply anymore.

samdealy: I think that using MobUri is confusing because that's the name of a [class](https://github.com/mobilecoinofficial/android-sdk/blob/266edb798c6d9a09d03e3836a23cb2b39065a57c/android-sdk/src/main/java/com/mobilecoin/lib/network/uri/MobUri.java). When I first read this, I thought you referring to that class. 

I am in favor of using "T" because it's a well-established convention. When I see "T" I know that it's a generic. That said, if you want to express that the "T" must extend "MobileCoinUri" it's probably best to make 'MobileCoinUri" an interface. That way, you wouldn't need to use generics and can just type the param as "MobileCoinUri". 

If that change will take too much time, I'd recommend either using "T" or some other name that makes it clear it's a generic and not the MobUri class that exists.

voloshyn: Let's discuss the naming offline.
I forgot we have MobUri, I'll rename that to something else.

voloshyn: I've removed the generics

samdealy: glynn

samdealy (approved):
LGTM

PR #36, opened 2021-07-13
Reset attested state on erros

voloshyn:
### Motivation

An attested state may become invalid after errors, so it has to be reset.

### In this PR
* Reset attested state on errors
* Reset transport with invalid attestation

### Future Work
* Simplify re-attestation by abstracting out error processing


android-sdk/src/main/java/com/mobilecoin/lib/AttestedViewClient.java
```diff
@@ -119,6 +119,7 @@ protected synchronized void attest(@NonNull Transport transport)
         Attest.Message message = encryptMessage(requestBuilder.build(), aadRequestBuilder.build());
```
eranrund, line 119: This is outside of the retry loop, so it retrying will not work.

samdealy (approved):
LGTM

eranrund (changes requested):

eranrund (approved):
LGTM once CI is happy

PR #37, opened 2021-07-15
Decommission rng

samdealy:
### Motivation

The SDK uses an RNG supplied by Fog Ingest to generate private txo search keys. The RNG seeds are specific to a given enclave, so when we upgrade to a new enclave the RNGs associated with the old enclave will not produce new search keys. These old RNGs are said to be "decommissioned," and the SDK should no longer generate search keys from them once we've gotten the last txo associated with the seed. 

The SDK didn't necessarily need to implement this functionality up until this point because there weren't a lot of decommissioned RNGs. However, as txos grow over time, querying decommissioned RNGs would negatively impact performance for no benefit to the user. As such, we're implementing this behavior now.

### In this PR
Adds a field to FogSeed called "isObsolete" that signifies if the seed is (a) decommissioned and (b) has had all of its txos retrieved. If the FogSeed is obsolete, then we won't use it to make Fog ingest queries.

Fixes [Decommission RNGs on Android](https://app.asana.com/0/1200318482173758/1200318482173797/f)

### Testing
Eran ran the Fog conformance tests, and confirmed that the tests passed. I also added logs in the new logic branches, and Eran confirmed that he saw those logs, which indicates that the new logic is being executed. 

### Future Work
* Non goal: Fix attestation issue
* Refactor TxOutStore logic to be more modular.
* Implement Parcelable serialization instead of Serializable

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
@@ -124,15 +124,15 @@ final class FogSeed implements Serializable {
         return utxos;
     }
 
-    public long getIngestInvocationId() {
+    long getIngestInvocationId() {
         return ingestInvocationId;
     }
 
-    public boolean isObsolete() {
+    boolean isObsolete() {
         return isObsolete;
     }
 
-    public void markObsolete() {
+    void markObsolete() {
         isObsolete = true;
     }
 
--- a/android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
@@ -29,8 +29,8 @@ final class FogSeed implements Serializable {
     // Data that comes straight from fog.
     private byte[] nonce;
     private int rngVersion;
-    // True if the seed is decommissioned and all utxos have been retrieved.
-    private boolean isDeprecated;
+    // True if the seed is (a) decommissioned and (b) all utxos have been retrieved.
+    private boolean isObsolete;
     private UnsignedLong startBlock;
     private ArrayList<OwnedTxOut> utxos;
 
@@ -128,12 +128,12 @@ final class FogSeed implements Serializable {
         return ingestInvocationId;
     }
 
-    public boolean isDeprecated() {
-        return isDeprecated;
+    public boolean isObsolete() {
+        return isObsolete;
     }
 
-    public void deprecate() {
-        isDeprecated = true;
+    public void markObsolete() {
+        isObsolete = true;
     }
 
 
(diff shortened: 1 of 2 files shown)


voloshyn (changes requested):
All public methods need a javadoc. Although see the comment below, maybe they don't need to be public.

android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
```diff
@@ -119,6 +124,19 @@ public UnsignedLong getStartBlock() {
         return utxos;
     }
 
+    public long getIngestInvocationId() {
```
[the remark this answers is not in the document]
samdealy: They don't need to be public. Changed them to package-private.

android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
```diff
@@ -22,11 +22,15 @@
 
     // Bump serial version and read/write code if fields change
     private static final long serialVersionUID = 1L;
+
+    private final long ingestInvocationId;
```
eranrund, line 26: I might be totally wrong here, but I think this needs to get serialized in `writeObject` and `readObject`, which would also imply a format change and perhaps a potential issue to existing SDK users?

android-sdk/src/main/java/com/mobilecoin/lib/TxOutStore.java
```diff
@@ -46,6 +44,8 @@
 
     // A map of nonce -> Seed.
     private HashMap<Integer, FogSeed> seeds;
+
+    private Set<Long> decommissionedIngestInvocationIds;
```
eranrund, line 48: Similar to my other comment, it looks like there is work that needs to happen to persist this and deal with versions that don't have it.

voloshyn (approved):

android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
samdealy: I made the changes to readObject and writeObject. 

I don't think that this change is an issue for existing SDK users because, to my knowledge, the serialization / deserialization feature is not currently being used by any of our clients. In fact, IIRC, this Serialization code is a placeholder for when apps will use it, and we're not even going to be using the Serialization class- instead will use [Parcelable][https://developer.android.com/reference/android/os/Parcelable). 

@voloshyn Can you confirm that this change is backwards compatible and won't break clients?

android-sdk/src/main/java/com/mobilecoin/lib/TxOutStore.java
samdealy: Same as above.

android-sdk/src/main/java/com/mobilecoin/lib/FogSeed.java
voloshyn: Yes, we are about to remove it in favor of Parcelable, this is not going to break clients

eranrund (approved):

PR #38, opened 2021-08-09
Serialization

samdealy:
### Motivation
For every host app session, the SDK makes a series of calls to the Fog View service to retrieve Random Number Generator (RNG) seeds and TxOuts for the user. The first request is always for just RNGs alone, and then subsequent requests use the retrieved RNGs to retrieve all of the user's TxOuts. 

This is a  time-consuming process, especially if the user has a lot of TxOuts, because we have to make this initial request each time, and then we have to wait for the Fog View server to find and return a lot of RNGs and TxOuts. 

### In this PR
To remove this initial "sole RNG" request and not have all of the user's TxOuts returned for every app session, this PR implements cacheing for the TxOut store. This class contains all of the user's RNG seeds, TxOuts, and other Fog View metadata, so we won't need to make this initial request and retrieve all the user's TxOuts.

Note that I had to make one change to how AttestViewClient works. Previously, that class stored the lastKnownEventId and lastBlockIndex as fields, and attached these values to each Fog view query. Now that we want to serialize SDK state, I moved these fields to the TxOutStore. Please double check that this can be done safely.

Also, in order to make TxOutStore serializable, I made ClientKexRNG serializable. This class implements native code and doesn't encapsulate any fields, so I opted to serialize it by using the native code to serialize and deserialize a ClientKexRNG from a StoredRng using protobuf serialization. A forthcoming PR in the fog repo will provide the source code for these changes to the native binary.

Asana [Ticket](https://app.asana.com/0/1200318482173758/1200616675344130/f) 

### Test

- Integration test in MobileCoinClientTest

### Future Work
* Migrate from Serializable to Parcelable interface (PR will be posted once this PR is merged).
* Fog PR that adds source code for native lib changes related to Java ClientKexRNG class.


android-sdk/src/main/java/com/mobilecoin/lib/TxOutStore.java
```diff
@@ -37,7 +37,7 @@
 import java.util.stream.Collectors;
 
 final class TxOutStore implements Serializable {
-    private static final String TAG = TxOutStore.class.getName();
+     static final String TAG = TxOutStore.class.getName();
```
voloshyn, line 40: * why was the visibility changed?
* lint: extra space before static

voloshyn (approved):
LGTM after the comments are fixed

android-sdk/src/main/java/com/mobilecoin/lib/TxOutStore.java
dolanbernard: The visibility change might be a question for Sam. I checked usages and it doesn't seem to be used outside of TxOutStore.java. So maybe he meant to change it back. Setting it as private again doesn't seem to break anything, so if the build passes I'll merge.

PR #44, opened 2021-09-15
Release 1.2.0-pre0

samdealy:
### In this PR
This is the PR for the 1.2.0-pre0 release. See the CHANGELOG.MD file for what's included in this release.


the-real-adammork:
I don't see a submodule change where do the binaries come from ?

samdealy:
> I don't see a submodule change where do the binaries come from ?

We don't use a submodule. Instead, we manually copy over the jnILibs and proto from the binaries / protos built from the mobilecoin repo.

voloshyn (approved):

PR #47, opened 2021-10-06
Implement internal load balancer

samdealy:
### Motivation
Right now, it's difficult for clients to communicate with multiple consensus nodes. If a consensus node goes down and they'd like to communicate with another node, they have to reinstantiate a `MobileCoinClient` and supply a new `consensusUri`. 

### In this PR
* Creates a `LoadBalancer` interface, which supplies a URI.
* Creates `RandomLoadBalancer`, which implements `LoadBalancer` and supplies a random URI from a list of URIs.
* Creates a new `MobileCoinClient` constructor that allows clients to pass in a list of `consensusUris`. Refactors tests to use this constructor
* Uses `RandomLoadBalancer` to provide a `consensusUri` for client classes that communicate with consensus nodes. This URI is used once per attestation session.
* Nulls out the `restClient` on shutdown. This ensures that a new `consensusUri` is chosen for the next attestation session.

Fixes this [ticket](https://app.asana.com/0/1200318482173758/1200761120709344/f).

### Testing
I created a local test that experiences a `NetworkException` when a transaction is submitted. Attestation resets, and then I submit another transaction. I verified that a new `consensusUri` is used for the second transaction submission. I tested this for both GRPC and HTTP clients.

### Future Work
* Investigate whether or not we should use `LoadBalancer` for Fog URIs.
* Add integration tests that emulate the testing I outlined in the Testing section.

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
@@ -24,12 +24,14 @@ import com.mobilecoin.lib.log.Logger;
 import com.mobilecoin.lib.network.TransportProtocol;
 import com.mobilecoin.lib.network.uri.ConsensusUri;
 import com.mobilecoin.lib.network.uri.FogUri;
+import com.mobilecoin.lib.network.uri.MobileCoinUri;
 import com.mobilecoin.lib.util.Result;
 import com.mobilecoin.lib.util.Task;
 import consensus_common.ConsensusCommon;
 import fog_ledger.Ledger;
 import java.math.BigInteger;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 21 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -102,20 +104,47 @@ public MobileCoinClient(
...
+    ) throws InvalidUriException {
+        this(accountKey, fogUri, Collections.singletonList(consensusUri), clientConfig);
+    }
+
+    /**
+     * Construct new {@link MobileCoinClient} instance
+     * If the service URIs don't specify the ports explicitly, port 443 will be used by default.
+     *
+     * @param accountKey   user's accountKey
+     * @param fogUri       a complete URI for the fog service
```
the-real-adammork, line 116:
```suggestion
     * @param fogUri       a complete URI for the fog service
```
```suggestion
     * @param fogUri       a complete URI's for the fog service
```

android-sdk/src/main/java/com/mobilecoin/lib/RandomLoadBalancer.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public MobileCoinUri getNextServiceUri() {
+    if (serviceUris.size() == 1) {
+      return serviceUris.get(0);
+    }
+
+    return getNextServiceUriForMultipleUris();
+  }
+
+  private MobileCoinUri getNextServiceUriForMultipleUris() {
```
the-real-adammork, line 46: was there any consideration to remove URI's (for a time) if it fails ?

the-real-adammork (approved):

PR #55, opened 2021-12-04
Add bip39 entropy support to TransferPayload

voloshyn:
### Motivation

Add bip39 entropy to the TransferPayload


dolanbernard (approved):
Was writing a comment but you fixed it before I could submit. Looks good if tests pass

the-real-adammork (approved):
looks good

PR #58, opened 2022-01-08
Test Short Fog URI

dolanbernard:
### Motivation
Creating a test for test net short fog uri

### In this PR
* Add test for short for uri
* Fixed an issue where MobileCoinClientBuilder would only take some values from the TestFogConfig after being set manually
* Finished MobileCoinClientBuilder

### DIFF
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java
@@ -23,6 +23,7 @@ public class TestFogConfig {
     public static final short FOG_REPORT_PRODUCT_ID = 4;
 
     private final Uri fogUri;
+    private final Uri shortFogUri;
     private final List<Uri> consensusUris;
     private final String username;
     private final String password;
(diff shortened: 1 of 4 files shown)


android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java
```diff
@@ -113,6 +122,7 @@ private static TestFogConfig getFogConfig(Environment.TestEnvironment testEnviro
                 "fog://fog.%s.mobilecoin.com",
                 testEnvironment.getName()
         ));
+        final Uri shortFogUri = Uri.parse(testEnvironment == Environment.TestEnvironment.TEST_NET ? "fog://fog-rpt-stg.namda.net" : "");
```
voloshyn, line 125: If this data is needed for only 1 test and it's valid for only 1 environment, adding it to TestFogConfig for all tests/environments creates confusion. Long URIs are used to initialize the clients, and short URI is used in the PublicAddress so clients know where to download the reports from. In this code, the short uri is ignored by all the tests except one. If the short URI is needed for only 1 test it's probably better to isolate the data within that scope.
Another option is to implement it to properly affect all the environments, but I don't see any gains from that at this point.

voloshyn (changes requested):

android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java
dolanbernard: Yeah, I was thinking about that. The reason why I put it here instead of just in the one test was in case we start using short URIs for other environments. But I can just move this to the one test and take it out of the config.

voloshyn (approved):

PR #62, opened 2022-01-22
Add CI configuration to automate publishing tagged releases

MCrank:
### Motivation

Publishing the Android-sdk to a Maven repository has been a manual process until now.  This PR adds CircleCI functionality to automate the publishing of the AAR file to cloudsmith for TAGGED releases of the Master branch in the format of `v.x.x.x*`

### In this PR
* Added a new CirclCI workflow and job for publishing the SDK for tagged releases on the master branch
* Added logic in publish.gradle to pull secrets from environment variables when running in CI or from the local.properties file if running locally
* Added new `publish` entry to makefile to handle needed steps for publishing to cloudsmith

### Future Work
* This work did not take into account building the android bindings  We should look at options to publish and version those dependencies which can be referenced here.
* Do we need to include updates to the readme for building the bindings and putting them in this repo?  If we can get the work in the above bullet working it would negate this issue.
* A possible optimization would be for us to create a base build image and just pull that image vs rebuilding it each time we run a step in the make file.  We could schedule this to get rebuilt periodically, i.e - weekly
   * This is not too major of a thing as it doesn't take long to build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,14 +57,12 @@ workflows:
       - test
 
   android-sdk-release:
-    triggers:
-      - schedule:
+    jobs:
+      - android-release
           filters:
             branches:
               only:
                 - ci
-    jobs:
-      - android-release
 
   # android-bindings-release:
   #   triggers:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
       - run:
           name: Store Google Service Account
           command: echo $GCLOUD_SERVICE_KEY > service-key.json
-      
+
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -37,9 +37,24 @@ jobs:
...
     jobs:
       - test
+  android-sdk-release:
+    jobs:
+      - android-release:
+          filters:
+            branches:
+              only:
+                - master
+            tags:
```
the-real-adammork, line 59: nice, Im gonna add this to iOS

voloshyn (approved):
This is great!

PR #64, opened 2022-01-27
Http build

dolanbernard:
### Motivation

Separating network code into separate source sets and creating flavors for GRPC enabled and HTTP only builds

### In this PR
* GRPC and HTTP build variants
* Source sets for GRPC and HTTP builds

### Future Work
* Updates to test running to cover all build variants


voloshyn (approved):
LGTM

PR #67, opened 2022-03-10
Recoverable Transaction History

dolanbernard:
### Motivation

This PR merges some very old Recoverable Transaction History (RTH) work by @samdealy  with the new android-sdk changes from the following months. In addition to merging, this PR includes various additions and updates.

### In this PR
* Merge for Sam's Recoverable Transaction History implementation from 09/2021
  * TxOutMemoBuilder
  * Sender and Destination Memos and MemoData
  * TxOutMemoParser
  * Bindings updates
  * RTH functionality tests
* New serialization code for RTH
* Refactors for RTH in order to remove private keys from RTH classes
* Updates to code and tests for larger RTH memo size
* Update to support token ID
* Update to TransactionBuilder to support different block versions
* Fix for one-time private key generation using default subaddress for change outputs
* Added tests for new and some existing RTH and serialization functionality

### Future Work
* Other 1.2 work unrelated to RTH
* Revisit prior idea of a higher-level RTH API with @voloshyn


android-sdk/src/main/java/com/mobilecoin/lib/TxOutMemo.java
```diff
@@ -0,0 +1,37 @@
...
+import android.os.Parcelable;
+
+import androidx.annotation.NonNull;
+
+/** Represents a TxOut memo. */
+public abstract class TxOutMemo extends Native implements Parcelable {
+
+  @NonNull
+  protected final TxOutMemoType memoType;
+  protected transient boolean validated = false;
```
dolanbernard, line 13: Transient keyword added here for documentation purposes, even though it doesn't have an effect on our new serialization. Putting this keyword here conveys that if a TxOutMemo is serialized and then deserialized, it will be considered invalid until verified, regardless of whether or not it was already validated before serialization. We could save this part of the state if we want to, but this functionality is consistent with the original RTH code before I made changes for serialization.

voloshyn (approved):
LGTM! 🚀
Nit: TransactionBuidler version could be an enum for readability

PR #68, opened 2022-03-16
Fog sync

dolanbernard:
### Motivation

After a Fog restart/redeploy, Fog will take some time to process available blocks. This can cause issues such as misreported balance to users during that time. To fix this, we can detect when Fog has fallen behind and throw an Exception with some information. Based on current data, we consider a block index discrepancy of about 10 to be exceptional. For now, we will use this threshold to detect a Fog sync and throw an Exception accordingly.

### In this PR
* Detection for Fog sync
* FogSyncException containing information on the status of Fog in the event it falls behind
* Unit tests to simulate different Fog states


android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -221,11 +222,17 @@ public AccountSnapshot getAccountSnapshot(UnsignedLong blockIndex) throws Networ
...
-                    fogBlockClient
-            );
+            try {
+                txOutStore.refresh(
+                        viewClient,
+                        ledgerClient,
+                        fogBlockClient
+                );
+            } catch(FogSyncException e) {
+                if(blockIndex.compareTo(e.getFogBlockIndex()) >= 0) {
```
voloshyn, line 232: Why block indexes check is needed after `FogSyncException` is already caught? Is it possible that it was thrown without a problem?
Also if this check is needed indeed, should it compare with `FOG_SYNC_THRESHOLD`?

android-sdk/src/main/java/com/mobilecoin/lib/exceptions/FogSyncException.java
```diff
@@ -0,0 +1,44 @@
+package com.mobilecoin.lib.exceptions;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.mobilecoin.lib.UnsignedLong;
+
+public class FogSyncException extends InvalidFogResponse {
```
voloshyn, line 8: please document what does this exception mean and in which cases it is thrown

android-sdk/src/main/java/com/mobilecoin/lib/exceptions/FogStatusException.java
```diff
@@ -0,0 +1,45 @@
+package com.mobilecoin.lib.exceptions;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.mobilecoin.lib.UnsignedLong;
+
+public class FogStatusException extends InvalidFogResponse {
```
voloshyn, line 8: please document what does this exception mean and in which cases it is thrown

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -221,11 +222,17 @@ public AccountSnapshot getAccountSnapshot(UnsignedLong blockIndex) throws Networ
...
+                    throw new InvalidFogResponse("Cannot create up-to-date snapshot until Fog sync finishes. Try again later.", e);
```
voloshyn, line 233: Since it's a fog sync error, I'd rather throw `FogSyncException` instead of `InvalidFogResponse`?

voloshyn (commented):
`FogSyncException` & `FogStatusException` seems to have very little difference, can we have just one `FogSyncException` instead?

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
dolanbernard: > Why block indexes check is needed after `FogSyncException` is already caught? Is it possible that it was thrown without a problem? Also if this check is needed indeed, should it compare with `FOG_SYNC_THRESHOLD`?

Let's say consensus is at block 30, fog is at block 20, and I want a snapshot of block 10. When refresh is called, it will throw the sync exception because fog is off by the threshold. But the txoutstore was still updated to block 20. Therefore I can say that a snapshot at block 10 is valid. This is probably a rare scenario, but I don't consider it a problem in this case. That is what this check is for.

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
dolanbernard: OK, I guess we can pass these all the way up.

dolanbernard:
> `FogSyncException` & `FogStatusException` seems to have very little difference, can we have just one `FogSyncException` instead?

There isn't really a reason why we cant, I just thought that maybe having the extra information could be helpful.

PR #69, opened 2022-03-23
Add Action to perform tests against specified network

MCrank:
### Motivation

This is needed to allow individuals the ability to perform tests against the specified network.  Primarily the infrastructure team will use this to validate any configuration changes do not break functionality.  Alleviates disruption to the mobile dev team to run out of band tests for the infrastructure team.This can also be used by anyone else needing to manually kick off a test.

### In this PR
* Added a GitHub action `workflow_dispatch` that allows authorized users to kick off the android-sdk tests by supplying which network to run the tests against.
* Added the necessary Action Secrets to perform the tests.
* Added a `timeout` value to the gcloud test cmdline.  Default is 15m and with the inclusion of additional tests we were hitting this limit on testnet.

### Follow up
* Look at optimizing tests to clean up technical debt.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,0 +1,52 @@
+name: sdk-tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      network:
+        description: 'MobileCoin Network'
+        required: true
+        default: 'MOBILE_DEV'
+        type: choice
+        options:
+          - 'MOBILE_DEV'
+          - 'ALPHA'
+          - 'TEST_NET'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: GCP Service Account
+        run: |
+          echo '${{ secrets.GCP_SERVICE_KEY }}' > "service-key.json"
+
+      - name: Provide Dev Credentials
+        env:
+          TEST_DEV_USER: ${{ secrets.TEST_DEV_USER }}
+          TEST_DEV_PASSWORD: ${{ secrets.TEST_DEV_PASSWORD }}
+        run: |
+          sed -i 's/REPLACE_TEST_DEV_USER_STRING/'"${TEST_DEV_USER}"'/g' android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java && \
+          sed -i 's/REPLACE_TEST_DEV_PASSWORD_STRING/'"${TEST_DEV_PASSWORD}"'/g' android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java
+
+      - name: Set the Network
+        env:
+          NETWORK_UNDER_TEST: ${{ github.event.inputs.network }}
+        run: sed -i 's/TestEnvironment.ALPHA/'"TestEnvironment.${NETWORK_UNDER_TEST}"'/g' android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
+
+      - name: Provide DevNet Entropies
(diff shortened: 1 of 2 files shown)


voloshyn (approved):

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          echo '${{ secrets.GCP_SERVICE_KEY }}' > "service-key.json"
+
+      - name: Provide Dev Credentials
+        env:
+          TEST_DEV_USER: ${{ secrets.TEST_DEV_USER }}
+          TEST_DEV_PASSWORD: ${{ secrets.TEST_DEV_PASSWORD }}
+        run: |
+          sed -i 's/REPLACE_TEST_DEV_USER_STRING/'"${TEST_DEV_USER}"'/g' android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java && \
+          sed -i 's/REPLACE_TEST_DEV_PASSWORD_STRING/'"${TEST_DEV_PASSWORD}"'/g' android-sdk/src/androidTest/java/com/mobilecoin/lib/TestFogConfig.java
```
jgreat, line 34: This works but is a bit messy, just put the regex in double quotes.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Set the Network
+        env:
+          NETWORK_UNDER_TEST: ${{ github.event.inputs.network }}
+        run: sed -i 's/TestEnvironment.ALPHA/'"TestEnvironment.${NETWORK_UNDER_TEST}"'/g' android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
```
jgreat, line 39: Longer term, can we specify the target urls.  Development environments will have dynamic names once we get to the new CI

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Provide DevNet Entropies
```
jgreat, line 41: This seems odd that there are 2 different ways to specify the wallet/key entropy

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: GCP Service Account
```
jgreat, line 24: ? what is this are we running things out of GCP?

jgreat (changes requested):

.github/workflows/tests.yaml
voloshyn: Integration tests require the Android runtime, the CI compiles them locally and then uploads the binaries to Firebase which can run them on any device matrix needed. The credentials are needed to access Firebase.

.github/workflows/tests.yaml
voloshyn: All networks used to have root-entropy based accounts, but it didn't support 24 words recovery, so we've upgraded to slip10. Dev nets haven't been updated yet, that's why two ways of supplying accounts.

.github/workflows/tests.yaml
voloshyn: Yes, let us know when, it would require a few simple changes on the java side.

.github/workflows/tests.yaml
MCrank: Cleaned up the commands. Also adjusted RegEx for the network as the value being checked was changed in a recent PR so that `sed` command would have been broken and is the setting pretty much for this whole PR.

jgreat (approved):

dolanbernard (approved):

PR #73, opened 2022-04-08
Block Version Update

dolanbernard:
### Motivation

Previously, as a placeholder, the block version was hard-coded. This PR dynamically updates the block version for TransactionBuilder from the LastBlockInfo response.

### In this PR
* Dynamically updating block version
* Some synchronization fixes for BlockchainClient


android-sdk/src/main/java/com/mobilecoin/lib/BlockchainClient.java
```diff
@@ -42,16 +42,23 @@
      * Fetch or return cached current minimal fee
      */
     @NonNull
-    synchronized UnsignedLong getOrFetchMinimumFee() throws NetworkException {
```
dolanbernard, line 45: getOrFetchLastBlockInfo is already synchronized, so we don't actually need this to be

android-sdk/src/main/java/com/mobilecoin/lib/TransactionBuilder.java
```diff
@@ -130,7 +130,7 @@ public Transaction build() throws TransactionBuilderException {
         } catch (Exception exception) {
             Logger.e(TAG, "Unable to set transaction fee", exception);
             throw new TransactionBuilderException(
-                    "Unable to build transaction from supplied " + "arguments");
+                    "Unable to build transaction from supplied arguments", exception);
```
dolanbernard, line 133: Unrelated to block version, but I realized before that this needed to be fixed

android-sdk/src/main/java/com/mobilecoin/lib/BlockchainClient.java
```diff
@@ -21,7 +21,7 @@
     private static final String TAG = BlockchainClient.class.getName();
     private static final BigInteger DEFAULT_TX_FEE = BigInteger.valueOf(10000000000L);
     private final Duration minimumFeeCacheTTL;
-    private ConsensusCommon.LastBlockInfoResponse lastBlockInfo;
+    private volatile ConsensusCommon.LastBlockInfoResponse lastBlockInfo;
```
dolanbernard, line 24: Don't store this response thread locally if it should be accessible from multiple threads

voloshyn (approved):

varsha888 (approved):

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutMemoIntegrationTest.java
```diff
@@ -45,6 +45,7 @@
...
 
+  private static final int MEMO_BLOCK_VERSION = 1;
```
varsha888, line 48: This is for my knowledge, why was the block version changed from 2 to 1?

PR #78, opened 2022-05-12
TransactionBuilder Shared Secret

dolanbernard:
### Motivation

This change is needed for transaction builder to expose a shared secret from add_output. This shared secret is necessary for Moby transaction history.

### In this PR
* TxOutContext, returned by TransactionBuilder addOutput and addChangeOutput


android-sdk/publish.gradle
```diff
@@ -1,6 +1,6 @@
 apply plugin: 'maven-publish'
 
-version '1.2.0-pre0'
```
the-real-adammork, line 3: is this correct ?

.circleci/config.yml
```diff
@@ -25,6 +25,10 @@ jobs:
       - run:
           name: Provide DevNet mnemonics
           command: echo $DEV_NET_MNEMONICS > android-sdk/src/androidTest/res/raw/dev_net_mnemonics
+          
+      - run:
```
the-real-adammork, line 29: was there an explicit ask for us to test with root entropies ? I can add it to iOS if there was

android-sdk/src/main/java/com/mobilecoin/lib/BlockchainClient.java
```diff
@@ -12,30 +12,28 @@
...
 import consensus_common.ConsensusCommon;
 
 class BlockchainClient extends AnyClient {
     private static final String TAG = BlockchainClient.class.getName();
     private static final BigInteger DEFAULT_TX_FEE = BigInteger.valueOf(10000000000L);
-    private final Duration minimumFeeCacheTTL;
+    private final long minimumFeeCacheTTLms;
     private volatile ConsensusCommon.LastBlockInfoResponse lastBlockInfo;
-    private LocalDateTime lastBlockInfoTimestamp;
+    private long lastBlockInfoTimestampms;
```
the-real-adammork, line 23:
```suggestion
    private long lastBlockInfoTimestamp_ms;
```
nitpick but `Timestampms` looks like a typo

the-real-adammork (approved):

.circleci/config.yml
dolanbernard: No, but the test keys for the android accounts on alpha use root entropies. A prior change I made preemptively updated both alpha and mobile dev to the new mnemonics when only dev net was updated. This is just to fix the alpha accounts until it gets updated to 1.2.

android-sdk/publish.gradle
dolanbernard: The version number that gets committed to the repo doesn't matter. Publishing is done from a local repository. I changed this to an invalid version number in case somebody forgets to change it before publishing. I want to avoid a situation where somebody publishes a new release and overwrites one that is being used. This helps prevent things like there being multiple different 1.2.0 clients "in the wild" or inadvertently changing somebody's build without knowledge or consent.

android-sdk/src/main/java/com/mobilecoin/lib/BlockchainClient.java
dolanbernard: Yeah, I wasn't sure what the best way to include the units in the name, but I guess that's it. I'll commit this suggestion and update the other names as well

PR #79, opened 2022-05-26
Adding getSharedSecret to OwnedTxOut

dolanbernard:
### Motivation

Moby needs access to the TxOut shared secret for transaction history

### In this PR
* Added getSharedSecret to OwnedTxOut

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
@@ -49,7 +49,6 @@ public class OwnedTxOut implements Parcelable {
     private final BigInteger value;
     private final RistrettoPublic txOutPublicKey;
     private final RistrettoPublic txOutTargetKey;
-    private final RistrettoPublic sharedSecret;
     private final byte[] keyImage;
     private int keyImageHash;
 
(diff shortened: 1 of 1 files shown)


voloshyn (changes requested):
Storing keys in cache is not safe, better to not cache this field at all. Instead, you could make `getSharedSecret` accept AccountKey and calculate it when needed.

android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
```diff
@@ -161,8 +160,8 @@ public RistrettoPublic getTargetKey() {
     }
 
     @NonNull
-    public RistrettoPublic getSharedSecret() {
-        return sharedSecret;
+    public RistrettoPublic getSharedSecret(AccountKey accountKey) throws TransactionBuilderException {
```
dolanbernard, line 163: @stevenspiel Will this be OK for you?

varsha888 (approved):

stevenspiel (approved):
LGTM

voloshyn (approved):

android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
voloshyn: Overall performance is better this way because you calculate `shared_secret` only when you need it instead of upon the creation of each `OwnedTxOut` instance. If this method is intended to be called multiple times we can keep the `sharedSecret` as a class field and return it upon subsequent queries, just skip that field when the object is serialized.

PR #87, opened 2022-06-16
Verify Correctness of Reconstructed Commitment

dolanbernard:
### Motivation

For TxOutRecords without commitment data, the commitment must be reconstructed. This PR computes the CRC of the reconstructed commitment and ensures it matches the CRC data in the TxOutRecord.

### In this PR
* Verification of reconstructed commitment data


voloshyn (approved):
LGTM

PR #90, opened 2022-06-30
Fix HttpRequester Auth

dolanbernard:
### Motivation

This PR fixes Basic authorization for the provided HttpRequester class

### In this PR
* Authorization fix for HttpRequester

### DIFF
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
@@ -6,7 +6,7 @@ import androidx.annotation.VisibleForTesting;
 
 @VisibleForTesting
 public class Environment {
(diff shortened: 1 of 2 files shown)


android-sdk/src/test/java/android/util/Base64.java
```diff
@@ -0,0 +1,13 @@
+package android.util;
+
+public class Base64 {
```
voloshyn, line 3: what is the purpose of this file?

dolanbernard: I added a dependency on the Android Base64 class to encode the credentials. The unit tests in HttpRequesterTest run outside the emulator though. So this class is added to mock the Base64 functionality needed for those tests to run.

voloshyn (approved):

PR #92, opened 2022-07-06
Unit tests and gradle update

varsha888:
### Motivation

Prior to this change, unit tests were run as instrumented tests which are usually slow. 
Robolectric library is used to run the unit tests in this PR. The library extends Android framework APIs and runs the code in sandbox. This lets us test each method in isolation and the tests are faster as it does not have to install/run on an emulator. 

### In this PR
Upgraded gradle to 7.2
Updated DockerFile to use gradle 7.2 JDK 11
Added roboelectric library
Added unit tests for AccountSnapshot

### Future Work
Add unit tests for other files
Use instrumented tests for testing a complete flow (e.g. submit transaction, view balance)

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
@@ -192,6 +192,11 @@ public final class MobileCoinClient implements MobileCoinAccountClient, MobileCo
         this.blockchainClient = blockchainClient;
     }
 
+    @NonNull
+    FogUntrustedClient getUntrustedClient() {
+        return untrustedClient;
+    }
+
     private List<MobileCoinUri> createNormalizedConsensusUris(List<Uri> consensusUris)
         throws InvalidUriException {
         List<MobileCoinUri> normalizedConsensusUris = new ArrayList<>();
@@ -925,7 +930,7 @@ public final class MobileCoinClient implements MobileCoinAccountClient, MobileCo
                 username,
                 password
         );
-        untrustedClient.setAuthorization(
+        getUntrustedClient().setAuthorization(
                 username,
                 password
         );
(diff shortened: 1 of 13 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -192,6 +192,11 @@ public MobileCoinClient(
         this.blockchainClient = blockchainClient;
     }
 
+    @NonNull
```
voloshyn, line 195: the package-private constructor allows the `untrustedClient` to be null on line 177

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -971,9 +976,7 @@ public synchronized void shutdown() {
         if (null != blockchainClient) {
             blockchainClient.shutdown();
         }
-        if (null != untrustedClient) {
-            untrustedClient.shutdown();
-        }
+        getUntrustedClient().shutdown();
```
voloshyn, line 979: `untrustedClient` can be null according to the package-private constructor, so this could crash

android-sdk/src/main/java/com/mobilecoin/lib/Native.java
```diff
@@ -2,12 +2,19 @@
...
     private final static String MOBILECOIN_LIB_NAME = "mobilecoin";
+    private static final String TAG = Native.class.toString();
 
     // Load JNI lib
     static {
-        System.loadLibrary(MOBILECOIN_LIB_NAME);
+        try {
+            System.loadLibrary(MOBILECOIN_LIB_NAME);
+        } catch (UnsatisfiedLinkError error) {
+            Logger.d(TAG, "Unable to load mobilecoin library");
```
voloshyn, line 16: I'd promote this error to `Logger.wtf` as this should never happen

voloshyn (approved):

PR #93, opened 2022-07-11
Add ProposeTxResult to InvalidTransactionException

dolanbernard:
### Motivation

This change is needed to make handling of different transaction errors easier. Prior to this, in order to handle different types of failures differently, String matching was required. With this change, enum value matching can be used.

### In this PR
* Add ProposeTxResult to InvalidTransactionException

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/exceptions/InvalidTransactionException.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/exceptions/InvalidTransactionException.java
@@ -5,24 +5,69 @@ package com.mobilecoin.lib.exceptions;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import com.mobilecoin.lib.Amount;
+import com.mobilecoin.lib.DefragmentationDelegate;
+import com.mobilecoin.lib.PendingTransaction;
+import com.mobilecoin.lib.PublicAddress;
+import com.mobilecoin.lib.Transaction;
+import com.mobilecoin.lib.TxOutMemoBuilder;
+
+import java.math.BigInteger;
+
 import consensus_common.ConsensusCommon;
 
+/**
+ * This class represents a {@link MobileCoinException} that results from a failed/rejected
+ * {@link com.mobilecoin.lib.Transaction} proposal.
+ *
+ * @see com.mobilecoin.lib.Transaction
+ * @see com.mobilecoin.lib.MobileCoinClient#prepareTransaction(PublicAddress, Amount, Amount, TxOutMemoBuilder)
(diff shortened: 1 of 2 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/exceptions/InvalidTransactionException.java
```diff
@@ -2,10 +2,75 @@
...
     }
+
+    /**
+     * Returns the {@link ConsensusCommon.ProposeTxResult} that caused this {@code InvalidTransactionException}.
+     *
+     * @return the failure reason for the {@link Transaction}
+     * @see ConsensusCommon.ProposeTxResult
+     * @since 1.2.2
+     */
+    @Nullable
```
voloshyn, line 71: This can be `@NonNull` if the deprecated constructor deleted

dolanbernard: It can and will be, but then we're removing a public API function. It could be in use somewhere, so maybe we should just deprecate it for now.

android-sdk/src/main/java/com/mobilecoin/lib/exceptions/InvalidTransactionException.java
```diff
@@ -2,10 +2,75 @@
...
+     *
+     * @param message the {@link Exception} message
+     * @deprecated Deprecated as of 1.2.2. Please use
+     * {@link InvalidTransactionException#InvalidTransactionException(ConsensusCommon.ProposeTxResult)}.
+     * @see InvalidTransactionException#InvalidTransactionException(ConsensusCommon.ProposeTxResult).
+     * @see ConsensusCommon.ProposeTxResult
+     * @since 1.0.0
+     */
+    @Deprecated
     public InvalidTransactionException(@Nullable String message) {
```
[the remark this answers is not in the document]
dolanbernard: We can't make the new one package private because `package com.mobilecoin.lib.exceptions` is a completely different package than `package com.mobilecoin.lib`, where we need to create it. So there must be a `public` way to create the exception unless we move all the exceptions.

stevenspiel (approved):
Looks good to me given Alex's suggestions.

android-sdk/src/main/java/com/mobilecoin/lib/exceptions/InvalidTransactionException.java
dolanbernard: We could just change the annotation and nothing else too. That decision is really a matter of whether we want new code to possibly have a false-positive null warning or old code to possibly have a false negative code warning. The former seems safer, in my opinion, but even with the latter, you avoid all problems by just not using a deprecated API.

PR #94, opened 2022-07-12
return consensus block index from submitTransaction

stevenspiel:
### Motivation

[task](https://www.pivotaltracker.com/story/show/182613898)

[comment on Moby Idempotence document](https://docs.google.com/document/d/1ayX6H_EjeCRVhu8lqUi7y1-D14RqyGDrewu70XcpYGY/edit?disco=AAAAbNCWhqI):

> Alex Voloshyn: 
> 1. KeyImages can be retrieved from TxOuts. (i.e. OwnedTxOut.getKeyImage)
> 2. Transaction object contains all KeyImages as a set (i.e. transaction.getKeyImages)
> 3. Fog View, Fog Ledger, and Consensus are all separate services with separate block heights. The SDK simplifies working with Fog by synchronizing View and Ledger block heights, but not consensus. So if you see that consensus is at block 1000 and it returns ContansSpentImages you will need to query Fog until it reports block height 1000 as well. (getTransactionStatus returns Status that contains current Fog block index, AccountSnapshot contains block index, etc). **SubmitTransaction does not return the consensus block index, so we will need to either add that or make BlockchainClient block height public to have it available at all times, not only after you submit**.

### In this PR
* return consensus block index from `submitTransaction`

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
@@ -578,7 +578,7 @@ public final class MobileCoinClient implements MobileCoinAccountClient, MobileCo
             Util.logException(TAG, invalidTransactionException);
             throw invalidTransactionException;
         }
-        return txResponse.getBlockCount();
+        return txResponse.getBlockCount() - 1;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


voloshyn (approved):
LGTM

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -578,6 +578,7 @@ public void submitTransaction(@NonNull Transaction transaction)
             Util.logException(TAG, invalidTransactionException);
             throw invalidTransactionException;
         }
+        return txResponse.getBlockCount();
```
dolanbernard, line 581:
```suggestion
        return txResponse.getBlockCount() - 1;
```
This returns block count. To make this the same as iOS, we should return the index (count - 1)

dolanbernard (changes requested):

dolanbernard (approved):

PR #95, opened 2022-07-12
make sure block count is never below zero

stevenspiel:
### Motivation

The Swift SDK ensures that the returning block index is [never below 0](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/master/Sources/Transaction/TransactionSubmitter.swift#L38). This does the same for Android.

### In this PR
* If `blockIndex` is 0, don't subtract 1 from it, making it negative.

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
@@ -568,8 +568,8 @@ public final class MobileCoinClient implements MobileCoinAccountClient, MobileCo
                 "transaction:", transaction);
         ConsensusCommon.ProposeTxResponse txResponse =
                 consensusClient.proposeTx(transaction.toProtoBufObject());
-        final long blockCount = txResponse.getBlockCount() > 0 ? txResponse.getBlockCount() - 1L : 0;
-        this.txOutStore.setConsensusBlockIndex(UnsignedLong.fromLongBits(blockCount));
+        final long blockIndex = txResponse.getBlockCount() > 0 ? txResponse.getBlockCount() - 1L : 0;
+        this.txOutStore.setConsensusBlockIndex(UnsignedLong.fromLongBits(blockIndex));
         ConsensusCommon.ProposeTxResult txResult = txResponse.getResult();
         int code = txResult.getNumber();
         if (0 != code) {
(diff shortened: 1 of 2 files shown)


voloshyn (approved):

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -568,7 +568,8 @@ public long submitTransaction(@NonNull Transaction transaction)
                 "transaction:", transaction);
         ConsensusCommon.ProposeTxResponse txResponse =
                 consensusClient.proposeTx(transaction.toProtoBufObject());
-        this.txOutStore.setConsensusBlockIndex(UnsignedLong.fromLongBits(txResponse.getBlockCount() - 1L));
+        final long blockCount = txResponse.getBlockCount() > 0 ? txResponse.getBlockCount() - 1L : 0;
+        this.txOutStore.setConsensusBlockIndex(UnsignedLong.fromLongBits(blockCount));
```
dolanbernard, line 572:
```suggestion
        final long blockIndex = txResponse.getBlockCount() > 0 ? txResponse.getBlockCount() - 1L : 0;
        this.txOutStore.setConsensusBlockIndex(UnsignedLong.fromLongBits(blockIndex));
```
This is good, but let's rename the variable here to blockIndex because it represents not the count, but the index

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -578,7 +579,7 @@ public long submitTransaction(@NonNull Transaction transaction)
             Util.logException(TAG, invalidTransactionException);
             throw invalidTransactionException;
         }
-        return txResponse.getBlockCount() - 1;
+        return blockCount;
```
dolanbernard, line 582:
```suggestion
        return blockIndex;
```
For the same reason as above

dolanbernard (changes requested):

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
stevenspiel: Ah, that's good to know.

dolanbernard (approved):
LGTM

PR #96, opened 2022-07-14
Add rng seed

stevenspiel:
### Motivation

Ensure idempotence by providing the random number generator to guarantee the same generated public key, which can only exist on the block chain at most one time.

### In this PR
* Adds a new constructor that includes the `rngSeed` param and passes it along to the rust bridge.

### Future Work
* Test that it works


dolanbernard:
> ### Future Work
> 
>     * Test that it works

Does this work? We should probably test before merging. It looks like there are build issues. Why is this change needed?

voloshyn:
Implemented here: https://github.com/mobilecoinofficial/android-sdk/pull/101, closing.

PR #111, opened 2022-09-23
Changing order of prepareTransaction RNG calls

dolanbernard:
### Motivation

This change is needed to ensure the public key of Transaction output is repeatable using any input(s) with the same RNG and seed

### In this PR
* Changed RNG calls in MobileCoinClient.prepareTransaction

### DIFF
--- a/android-sdk/build.gradle
+++ b/android-sdk/build.gradle
@@ -128,6 +128,6 @@ dependencies {
     androidTestImplementation deps.testing.rules
     androidTestImplementation deps.testing.espresso_core
     androidTestImplementation deps.testing.support_rules
-    androidTestImplementation deps.testing.okhttp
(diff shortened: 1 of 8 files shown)


the-real-adammork (approved):

voloshyn (approved):
LGTM. Nit for the future: let's create a test data repository. Hardcoding data inline makes it hard to read.
e.g. `final ChaCha20Rng rng = ChaCha20Rng.fromSeed(new byte[] {103, 111, 116, 111, 32, 104, 116, 116, 112, 115, 58, 47, 47, 98, 117, 121, 46, 109, 111, 98, 105, 108, 101, 99, 111, 105, 110, 46, 99, 111, 109, 0});
`

PR #116, opened 2022-12-08
RTH Payment ID Memos

dolanbernard:
### Motivation

#67 added SenderWithPaymentRequestMemo. This memo allows recipients of a Transaction to know whether or not a received TxOut corresponds to a payment request. There is currently no way for the user sending that Transaction to recover this information using RTH or any other data on the blockchain. This PR adds a new memo type: DestinationWithPaymentRequestMemo, which the user sends themselves via a change TxOut when sending a Transaction with a SenderWithPaymentRequestMemo.

This PR also adds Sender and DestinationWithPaymentIntentMemo. These memos are structured the same but differ in memo type. Having a distinct type for payment intent memos allows clients to implement a payment intent flow.

### In this PR
* DestinationWithPaymentRequestMemo with MemoData
* SenderWithPaymentIntentMemo with MemoData
* DestinationWithPaymentIntentMemo with MemoData
* Sender/Destination With Payment Intent MemoBuilders

### DIFF
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
@@ -6,7 +6,7 @@ import androidx.annotation.VisibleForTesting;
 
 @VisibleForTesting
 public class Environment {
-    public static final TestEnvironment CURRENT_TEST_ENV = TestEnvironment.TEST_NET;
+    public static final TestEnvironment CURRENT_TEST_ENV = TestEnvironment.MOBILE_DEV;
 
     static public TestFogConfig getTestFogConfig() {
         return TestFogConfig.getFogConfig(CURRENT_TEST_ENV);
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
@@ -6,7 +6,7 @@ import androidx.annotation.VisibleForTesting;
 
 @VisibleForTesting
 public class Environment {
-    public static final TestEnvironment CURRENT_TEST_ENV = TestEnvironment.MOBILE_DEV;
+    public static final TestEnvironment CURRENT_TEST_ENV = TestEnvironment.TEST_NET;
 
     static public TestFogConfig getTestFogConfig() {
         return TestFogConfig.getFogConfig(CURRENT_TEST_ENV);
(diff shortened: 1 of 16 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/DestinationMemo.java
```diff
@@ -9,9 +9,9 @@
 import java.util.Objects;
 
 /** Represents a destination memo, which corresponds to the "DestinationMemo" specification. */
-public final class DestinationMemo extends TxOutMemo {
```
voloshyn, line 12: Why not keep it final?

dolanbernard: Initially, DestinationWithPaymentRequestMemo and DestinationWithPaymentIntentMemo were going to extend DestinationMemo. The goal was to let clients only handle generic Destion and Sender memos.

The issue with this is that it doesn't work well with how the original RTH spec was finalized. We can't just get the memo data from the memo class. Each memo type has it's own memo data class and an appropriately named method to access it. So if we just make, for example DestinationWithPaymentIntentMemoData extend DestinationMemoData, then DestinationWithPaymentIntentMemo would need to provide two accessors: getDestinatonWithPaymentIntentMemoData() and getDestinationMemoData() which would both return the same object. This is unnecessarily confusing. I think this behavior would best be implemented with a new high-level API spec.

I can change this and all the other memo types to final, if we don't need them to be extended. But I don't know if there isna reason they need to be.

voloshyn: Not having `final` signals to the SDK users that the author intends for them to extend this class to work with it. If that's not your intention, then make it `final`. Also, simple overrides can have serious side effects, and non-final classes must take that into account, so unless you expect users to extend it, the safer way is to declare it `final`. (e.g. a private method calls a public method in the base class, then overrides the public method, etc)

android-sdk/src/androidTest/java/com/mobilecoin/lib/Environment.java
```diff
@@ -6,7 +6,7 @@
 
 @VisibleForTesting
 public class Environment {
-    public static final TestEnvironment CURRENT_TEST_ENV = TestEnvironment.TEST_NET;
+    public static final TestEnvironment CURRENT_TEST_ENV = TestEnvironment.MOBILE_DEV;
```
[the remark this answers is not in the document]
dolanbernard: Changed back to test net

voloshyn (approved):

PR #117, opened 2022-12-08
Only Fail RngTest on Multiple Repeat Values

dolanbernard:
### Motivation

RngTest sometimes causes random failures if it happens to generate a duplicate number (usually byte). This can cause the entire CircleCI workflow to need to be re-run numerous times to merge a PR. This PR should decrease the likelihood of random failures by reducing the number of randomly generated values.

### In this PR
* Reduce the number of randomly generated values in RngTest

### DIFF
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/RngTest.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/RngTest.java
@@ -6,6 +6,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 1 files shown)


voloshyn:
It would be better to modify the test to eliminate CI failures; reducing the likelihood only makes it less annoying. Please elaborate on why duplicate values cause build failures. If that's by design, maybe you could use a HashMap to eliminate duplicates.

dolanbernard:
> It would be better to modify the test to eliminate CI failures; reducing the likelihood only makes it less annoying. Please elaborate on why duplicate values cause build failures. If that's by design, maybe you could use a HashMap to eliminate duplicates.

We are using a HashMap to detect duplicates. I can rewrite the check for count of duplicates to try and detect repeat sequences rather than repeat values.

voloshyn (approved):

PR #118, opened 2022-12-08
Fix Insecure Transport Using TLS

dolanbernard:
### Motivation

If an insecure protocol is specified in Fog or Consensus URI, TLS should not be used. When the SSL socket factory is set in the managed channel builder, the manager channel builder automatically enables TLS. This PR moves the secure/insecure protocol check after the SSL socket factory specification. This will prevent the manager channel from using TLS if an insecure protocol is specified in the URI.

### In this PR
* Fix using TLS when insecure protocol is specified by URI

### DIFF
--- a/android-sdk/src/grpc/java/com/mobilecoin/lib/network/services/transport/grpc/GRPCTransport.java
+++ b/android-sdk/src/grpc/java/com/mobilecoin/lib/network/services/transport/grpc/GRPCTransport.java
@@ -42,16 +42,16 @@ public class GRPCTransport extends Transport {
                             uri.getUri().getHost(),
                             uri.getUri().getPort()
                     );
-            if (uri.isTlsEnabled()) {
-                managedChannelBuilder.useTransportSecurity();
-            } else {
-                managedChannelBuilder.usePlaintext();
-            }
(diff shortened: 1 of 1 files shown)


android-sdk/src/grpc/java/com/mobilecoin/lib/network/services/transport/grpc/GRPCTransport.java
```diff
@@ -42,17 +42,17 @@ public GRPCTransport(@NonNull MobileCoinUri uri, @NonNull ClientConfig.Service s
                             uri.getUri().getHost(),
                             uri.getUri().getPort()
                     );
-            if (uri.isTlsEnabled()) {
-                managedChannelBuilder.useTransportSecurity();
-            } else {
-                managedChannelBuilder.usePlaintext();
-            }
             Set<X509Certificate> trustRoots = serviceConfiguration.getTrustRoots();
```
voloshyn, line 45: SSLSocketFactory is not useful if TLS is disabled, can this go inside this check `if (uri.isTlsEnabled())`

dolanbernard: Yes, that works too

voloshyn (approved):

PR #125, opened 2023-02-07
Improve Account Creation API

dolanbernard:
### Motivation

There are three ways `AccountKey`s can be created:
1. `AccountKey.createNew`
2. `AccountKey.fromBip39Entropy`
3. `AccountKey.fromMnemonicPhrase`

The first of these is somewhat problematic. When an `AccountKey` is created using `AccountKey.createNew`, the `AccountKey` is created randomly. It doesn't take any recovery info passed into the API. However, after the `AccountKey` has been created, the `RootIdentity` cannot be recovered from it. So no recovery information can be extracted from the newly created `AccountKey`. So, short of just serializing the keys and storing the bytes somewhere, **_there is no way to create the same_** `AccountKey` **_twice using this API method._** Furthermore, this strategy of serializing the `AccountKey` is prone to storing insecurely, as it is not easy for people to record or reproduce.

The second is a bit better, as the recovery information is passed into the API, meaning the API caller must have the info before calling `AccountKey.fromBip39Entropy` and has the opportunity to save it first. However, a Bip39 entropy is still difficult for people to work with.

The third method provides the best user experience while making it more clear to developers that `AccountKey` recovery information should be saved securely _before_ it is created. And to make the process of creating a mnemonic easier for developers, a new method (`Mnemonics.createRandomMnemonic`) was added to assist with making cryptographically secure, random mnemonics. The old method of doing this would have been to create a Bip39 entropy manually and then use `Mnemonics.fromBip39Entropy`. But this isn't obvious, requires more code than should be necessary, and leaves it up to developers to securely generate the random account.

### In this PR
* Deprecate `AccountKey.createNew`
* Created new method for creating random mnemonics

### Future Work
* In a later version, remove deprecation tags and mark deprecated methods package-private as they are still useful for SDK testing purposes
* Update developer docs/integration guide to reflect new account creation process

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
       - run:
           name: make tests
           command: make tests
-          no_output_timeout: 30m
+          no_output_timeout: 45m
 
       - store_artifacts:
           path: android-sdk/build/reports
--- a/android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
+++ b/android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountTest.java
@@ -300,6 +300,23 @@ public class AccountTest {
         }
     }
 
+    @Test
+    public void testCreateAccountFromRandomMnemonic() throws Exception {
+        for(int i = 0; i < 100000; i++) {
+            Mnemonics.createRandomMnemonic();
+        }
+        for(int i = 0; i < 1000; i++) {
+            final String mnemonic = Mnemonics.createRandomMnemonic();
+            AccountKey.fromMnemonicPhrase(
+                    mnemonic,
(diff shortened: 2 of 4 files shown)


voloshyn (commented):
How do I create an account if I want to supply my own entropy?

voloshyn (approved):
Please update the PR description and then LGTM

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - run:
           name: make tests
           command: make tests
-          no_output_timeout: 30m
+          no_output_timeout: 45m
```
dolanbernard, line 36: This branch does not have the optimized test coverage yet. I will revert after merging.

PR #127, opened 2023-02-10
Crypto Box Encrypt

dolanbernard:
### Motivation

This PR adds asymmetric encryption features to the public API. This will allow applications to encrypt data using an account's public key and then decrypt it using the account's private key.

### In this PR
* Added `VersionedCryptoBox.versionedCryptoBoxEncrypt`
* Switched accessors for `AccountKey` keys to public

### DIFF
--- a/android-sdk/versions.gradle
+++ b/android-sdk/versions.gradle
@@ -28,7 +28,7 @@ versions.network = network_versions
 
 // JNI
 def jni_versions = [:]
-jni_versions.mobilecoin = '3.0.0-pre6'
+jni_versions.mobilecoin = '2.0.1-pre0'
 versions.jni = jni_versions
 
 // Testing
--- a/android-sdk/versions.gradle
+++ b/android-sdk/versions.gradle
@@ -28,7 +28,7 @@ versions.network = network_versions
 
 // JNI
 def jni_versions = [:]
-jni_versions.mobilecoin = '2.0.1-pre0'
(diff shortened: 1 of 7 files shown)


android-sdk/versions.gradle
```diff
@@ -28,7 +28,7 @@ versions.network = network_versions
 
 // JNI
 def jni_versions = [:]
-jni_versions.mobilecoin = '3.0.0-pre6'
+jni_versions.mobilecoin = '2.0.1-pre0'
```
voloshyn, line 31: Should this be 4.0.0.1? `versioned_crypto_box_encrypt` [was added in 4.0.0.1](https://github.com/mobilecoinofficial/android-bindings/pull/5/files)

dolanbernard: Yes, the version change is coming in the next PR for the release update. I just left this here to avoid the conflict in the next PR

voloshyn: Why would there be a conflict? This is a JNI dependency, the code in this PR won't work without it.

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - run:
           name: make tests
           command: make tests
-          no_output_timeout: 30m
+          no_output_timeout: 45m
```
dolanbernard, line 36: Will revert this after merge. This branch does not have test optimizations.

voloshyn (approved):

PR #129, opened 2023-02-14
Logging Update

dolanbernard:
### Motivation

This PR adds/improves SDK logging in critical areas of code. This will help debug issues in client apps and detect bugs. Logging was also removed or demoted in some places to reduce unimportant log output.

### In this PR
* Added/updated logging in areas of code most likely to fail
* Removed or demoted logging in areas that should never cause failures

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
@@ -789,6 +789,7 @@ public final class MobileCoinClient implements MobileCoinAccountClient, MobileCo
         try {
             Result<List<Ring>, Exception> ringsResult = ringsListFuture.get();
             if (ringsResult.isErr()) {
+                Logger.e(TAG, "Error fetching rings", ringsResult.getError());
                 // isError indicated that the error is non-null
                 throw Objects.requireNonNull(ringsResult.getError());
             } else if (ringsResult.isOk()) {
@@ -798,6 +799,7 @@ public final class MobileCoinClient implements MobileCoinAccountClient, MobileCo
             }
             Result<FogReportResponses, Exception> reportsResult = fogReportResponsesFuture.get();
             if (reportsResult.isErr()) {
+                Logger.e(TAG, "Error fetching reports", reportsResult.getError());
                 // isError indicated that the error is non-null
                 throw Objects.requireNonNull(reportsResult.getError());
             } else if (reportsResult.isOk()) {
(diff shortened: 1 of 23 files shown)


android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -1269,7 +1298,10 @@ List<Ring> getRingsForUTXOs(
 
         // Sanity
         if (realIndices.size() != utxos.size()) {
-            throw new IllegalStateException("BUG");
+            //throw new IllegalStateException("BUG");
+            final IllegalStateException ise = new IllegalStateException("BUG");
+            Logger.wtf(TAG, "Number of TxOut indices does not match number of TxOuts", ise);
```
carybakker, line 1303: Logger.wtf lol

android-sdk/src/main/java/com/mobilecoin/lib/SenderMemo.java
```diff
@@ -42,7 +42,7 @@ private SenderMemo(
       init_jni_from_memo_data(memoData);
     } catch(Exception e) {
       IllegalArgumentException illegalArgumentException =
-          new IllegalArgumentException("Failed to create an AccountKey", e);
+          new IllegalArgumentException("Failed to create SenderMemo", e);
```
carybakker, line 45: lots of old copy-paste errors in the logs, eh?  great to see these cleaned up

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -1219,18 +1237,27 @@ List<Ring> getRingsForUTXOs(
...
+                "utxo count:", utxos.size(),
+                "num txos in ledger", numTxOutsInLedger
+        );
         // Sanity check to ensure all UTXOs have unique indices
         HashSet<UnsignedLong> indices = new HashSet<>();
         for (OwnedTxOut utxo : utxos) {
             if (!indices.add(utxo.getTxOutGlobalIndex())) {
-                throw new IllegalStateException("utxos contains non-unique indices");
+                final IllegalStateException ise = new IllegalStateException("utxos contains non-unique indices");
+                Logger.e(TAG, "Encountered Exception while getting rings", ise);
```
carybakker, line 1249: Using "Encoutered" for the exception here seems strange since this is the throwing code (later code that catches exceptions also uses "Encountered" for the caught exceptions).  Maybe "Throwing Exception - error while getting rings", ise) or something like that for the all the throwing locations in this file?

carybakker (approved):
Left some feedback for the "Encountered" verbiage when throwing exceptions from MobileCoinClient versus catching exceptions...  will leave it up to you to change / leave as is as it's not critical.

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
dolanbernard: Hmm, yeah, this is an excellent point. In the logs it could look like two Exceptions occurred. I took your suggestion and re-worded it to say "throwing" here and in a couple other places.

Great find!

PR #138, opened 2023-11-16
Android SDK DCAP Attestation

dolanbernard:
### Motivation

This PR adds android-sdk client support for DCAP attestation. This will be needed for 6.0 MobileCoin networks.

### In this PR
* Support for DCAP attestation
* Legacy support for pre-6.0 EPID attestation
* Removing tech debt from some very old tests


android-sdk/src/main/java/com/mobilecoin/lib/Verifier.java
```diff
@@ -12,73 +12,44 @@
  * Verifier is used for attestation connection verification.
  */
 
-public final class Verifier extends Native {
-    private static final String TAG = Verifier.class.getName();
+public final class Verifier {
```
dolanbernard, line 15: The `Verifier` class is left in place but the underlying bindings are removed. The public API for the `Verifier` class is unaffected, but underneath, the `Verifier` class now uses the `TrustedIdentities` API. This prevents apps from requiring code changes immediately on upgrading. The old API still works as it did before, but it is marked as deprecated.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutMemoIntegrationTest.java
```diff
@@ -20,446 +23,202 @@
...
...669666d95bd2621d99d60506cb0d1a1e666f673a2f2f666f672e616c7068612e6d6f62696c65636f696e2e636f6d2aa60430820222300d06092a864886f70d01010105000382020f003082020a0282020100c853a8724bc211cf5370ed4dbec8947c5573bed0ec47ae14211454977b41336061f0a040f77dbf529f3a46d8095676ec971b940ab4c9642578760779840a3f9b3b893b2f65006c544e9c16586d33649769b7c1c94552d7efa081a56ad612dec932812676ebec091f2aed69123604f4888a125e04ff85f5a727c286664378581cf34c7ee13eb01cc4faf3308ed3c07a9415f98e5fbfe073e6c357967244e46ba6ebbe391d8154e6e4a1c80524b1a6733eca46e37bfdd62d75816988a79aac6bdb62a06b1237a8ff5e5c848d01bbff684248cf06d92f301623c893eb0fba0f3faee2d197ea57ac428f89d6c000f76d58d5aacc3d70204781aca45bc02b1456b454231d2f2ed4ca6614e5242c7d7af0fe61e9af6ecfa76674ffbc29b858091cbfb4011538f0e894ce45d21d7fac04ba2ff57e9ff6db21e2afd9468ad785c262ec59d4a1a801c5ec2f95fc107dc9cb5f7869d70aa84450b8c350c2fa48bddef20752a1e43676b246c7f59f8f1f4aee43c1a15f36f7a36a9ec708320ea42089991551f2656ec62ea38233946b85616ff182cf17cd227e596329b546ea04d13b053be4cf3338de777b50bc6eca7a6185cf7a5022bc9be3749b1bb43e10ecc88a0c580f2b7373138ee49c7bafd8be6a64048887230480b0c85a045255494e04a9a81646369ce7a10e08da6fae27333ec0c16c8a74d93779a9e055395078d0b07286f9930203010001";
```
dolanbernard, line 23: This old, hard-coded data was probably used because, at the time of writing these tests, no block version 1 networks were available for testing. This hard-coded data has needed to be updated multiple times since then and needed to be once again. This is a large but unnecessary source of tech debt. Since this is an integration test (and memos are already unit tested in mc core), there is no longer any reason why most of this data can't be fetched automatically.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutMemoIntegrationTest.java
```diff
@@ -20,446 +23,202 @@
...
-      sentTxOut = txOut1;
-      changeTxOut = txOut2;
-    } catch(Exception e) {
-      sentTxOut = txOut2;
-      changeTxOut = txOut1;
+    private static final int MEMO_BLOCK_VERSION = 1;
+
+    @Test
+    public void txOutMemoIntegrationTest() throws Exception {
+        final MobileCoinClient senderClient = MobileCoinClientBuilder.newBuilder().build();
```
dolanbernard, line 30: A single client is built and refreshed only once. After this, the transactions may be built to test memos without much additional overhead. So test performance isn't severely impacted.

android-sdk/src/main/java/com/mobilecoin/lib/FogReport.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public long getPublicKeyExpiry() {
+        Logger.i(TAG, "Getting public key expiry", null, publicKeyExpiry);
+        return publicKeyExpiry;
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        try {
+            finalize_jni();
```
eranrund, line 93: Is this missing a `if (rustObj != 0) {`?

eranrund (approved):

android-sdk/src/main/java/com/mobilecoin/lib/FogReport.java
dolanbernard: Good find! But that case should be handled by the catch block.

PR #142, opened 2023-12-15
Migrate Dependencies to AndroidX

dolanbernard:
### Motivation

This PR migrates android-sdk dependencies to AndroidX so Jetifier can be removed.

### In this PR
* Replace android.support dependencies with new androidx versions
* Remove Jetifier


the-real-adammork (approved):
looks good

PR #153, opened 2026-08-20
perf(SENTZ-5454): bound per-poll key-image scan with a start_block watermark

bcgreijnlautier:
The app polls balance every 10 seconds. Each poll checks every owned txo's key image against fog-ledger unconditionally. `fog/api/proto/ledger.proto`'s `KeyImageQuery.start_block` exists to avoid this. A client that already knows a key image was unspent as of block N can scan forward from there. Android never populated it. iOS's `FogKeyImageChecker` already does this with a per-key-image watermark.

This adds the same watermark on Android. `OwnedTxOut` gets a nullable `knownToBeUnspentBlockCount`, set only from an explicit `NotSpent` result carrying that exact key image, in `TxOutStore.updateTxOutsSpentState`. That is stricter than iOS, which also treats an omitted key image as unspent, because the proto attaches its unspent-through guarantee to a returned result. `AttestedLedgerClient.checkUtxoKeyImages` threads the watermark into `start_block`, 0 for a txo without one.

The field is not persisted through `OwnedTxOut`'s Parcelable implementation. A blob cached by one SDK version can be read back by another, and the two would disagree about the field set. Rebuilding the watermark costs one full scan after a process restart, always a safe fallback.

Worth knowing before this merges: the fog-ledger enclave at the pinned foundation revision `05cb699` discards `start_block`. `check_key_images` in `fog/ledger/enclave/impl/src/lib.rs` maps every query through `store.find_record(&key.key_image)`, a single ORAM point read with no block loop, so the client-side win only lands once fog honors the field.

`make tests` needs CI-injected mnemonics and isn't runnable locally, this PR relies on CI for it. `make build` passed locally, including the new unit tests.

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
@@ -42,11 +42,8 @@ public class OwnedTxOut implements Parcelable {
     private final Date receivedBlockTimestamp;
     private Date spentBlockTimestamp;
     private UnsignedLong spentBlockIndex;
-    // Highest block count fog-ledger has confirmed this key image unspent through.
-    // Scopes the next key-image scan to start_block; null means never checked.
-    // Not parcelled: Parcel has no per-object length prefix, so a field added here would
-    // misread trailing sibling bytes for every OwnedTxOut but the last in a cached blob.
-    // Resetting to null on process restart is always safe, it just forces one full scan.
+    // Highest block count fog-ledger confirmed this key image unspent through, null if never.
+    // Out of the parcel so a blob cached by another SDK version stays readable.
     @Nullable
     private UnsignedLong knownToBeUnspentBlockCount;
 
--- a/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
@@ -42,6 +42,13 @@ public class OwnedTxOut implements Parcelable {
     private final Date receivedBlockTimestamp;
     private Date spentBlockTimestamp;
     private UnsignedLong spentBlockIndex;
+    // Highest block count fog-ledger has confirmed this key image unspent through.
+    // Scopes the next key-image scan to start_block; null means never checked.
+    // Not parcelled: Parcel has no per-object length prefix, so a field added here would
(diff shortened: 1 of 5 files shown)


android-sdk/src/test/java/com/mobilecoin/lib/TxOutStoreWatermarkTest.java
```diff
@@ -0,0 +1,96 @@
...
+                        .setSpentAt(10)
+                        .setTimestamp(UnsignedLong.MAX_VALUE.longValue())
+                        .build())
+                .build();
+
+        Set<OwnedTxOut> queried = new HashSet<>(Arrays.asList(spentTxOut, unspentTxOut));
+        store.updateTxOutsSpentState(queried, response);
+
+        verify(spentTxOut, never()).setKnownToBeUnspentBlockCount(any());
+        verify(unspentTxOut).setKnownToBeUnspentBlockCount(UnsignedLong.fromLongBits(50));
```
stevenspiel: might be worth also testing that `checkUtxoKeyImages` reads `getKnownToBeUnspentBlockCount` and puts the value in that txo's `start_block`

android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
```diff
@@ -245,6 +253,15 @@ synchronized void setSpent(
         this.spentBlockTimestamp = spentBlockTimestamp;
     }
 
+    @Nullable
+    synchronized UnsignedLong getKnownToBeUnspentBlockCount() {
+        return knownToBeUnspentBlockCount;
+    }
+
+    synchronized void setKnownToBeUnspentBlockCount(@NonNull UnsignedLong knownToBeUnspentBlockCount) {
```
stevenspiel, line 258: `knownToBeUnspentBlockCount` is nullable. should we open a way for outside callers to reset this? I'm on the fence about it, but it caught my eye
```suggestion
    synchronized void setKnownToBeUnspentBlockCount(@Nullable UnsignedLong knownToBeUnspentBlockCount) {
```

stevenspiel (approved):
LGTM

android-sdk/src/main/java/com/mobilecoin/lib/OwnedTxOut.java
bcgreijnlautier: `spentBlockIndex` right above has the same pattern. A watermark only moves forward from a confirmed fog-ledger result, so no caller needs to clear it.

PR #154, opened 2026-08-20
ci(SENTZ-5455): migrate android-sdk CI from CircleCI to GitHub Actions

bcgreijnlautier:
### Motivation

CircleCI still runs CI while GitHub Actions holds dispatch-only copies of the same jobs. This finishes the move so CircleCI can be retired.

### In this PR
* `sdk-tests` runs on every branch push, matching the old `test` job. A push keeps the committed default network in `Environment.java`. Only a dispatch overrides it. Firebase Test Lab drives the tests, so `ubuntu-latest` needs only Docker and the GCP key.
* `Publish Android SDK` auto-publishes on a `v*` tag, matching the old `android-release` job. It has never run on Actions, so two traps are closed here. `make publish` drops `-t`, because a `run:` step has no terminal. A new step fails the job when the tag and `publish.gradle`'s version disagree. Without it, a tag cut without a bump republishes the old version.
* Deletes `.circleci/config.yml` and points the README badge at `sdk-tests`.
* Push is scoped to branches, so a `v*` tag triggers only the publish. The test step allows 60m, because `run_connected_tests.sh` runs two gradle assembles before Firebase's 45m timeout starts. The concurrency group is keyed on the ref, the event and the network, so a dispatch always finishes. The publish checkout stays on v3 to avoid forcing Node20 on the self-hosted runner.


.github/workflows/android-sdk-dispatch.yml
```diff
@@ -1,6 +1,9 @@
-name: Publish Android SDK 
+name: Publish Android SDK
 
 on:
+  push:
+    tags:
+      - 'v*'
   workflow_dispatch:
```
stevenspiel, line 7: this hasn't been run yet. it will likely succeed, and I'm ok with a follow up if something goes wrong, but something that caught my eye.

.github/workflows/tests.yaml
```diff
@@ -13,13 +16,17 @@ on:
           - 'ALPHA'
           - 'TEST_NET'
 
+concurrency:
+  group: sdk-tests-${{ github.ref }}
+  cancel-in-progress: true
```
stevenspiel: nit: allow for manually-kicked-off builds to complete

```suggestion
  cancel-in-progress: ${{ github.event_name == 'push' }}
```

stevenspiel (approved):
README still has the circleCI badge. Might be worth swapping that out.

Also Makefile:49 says

> Check if we are in the CircleCI environment

might want to change that to

> Check if we are in a CI environment

.github/workflows/android-sdk-dispatch.yml
bcgreijnlautier: That workflow has never run - `total_count` is 0. `make publish` no longer passes `-t`, since a `run:` step has no terminal. A new step fails the job when the tag and `publish.gradle`'s version disagree.

.github/workflows/tests.yaml
bcgreijnlautier: The group needed the event and the network too. `push` and `workflow_dispatch` share `github.ref`, so a push still cancelled a dispatch. A queued dispatch for a second network was displaced whatever `cancel-in-progress` said.

PR #156, opened 2026-08-25
ci(SENTZ-5470): Publish the android-sdk from a builder container

bcgreijnlautier:
### Motivation

The publish workflow has never succeeded. It moved from a CircleCI VM onto a containerized self-hosted runner. That runner has no `make`. The makefile publish target also wraps `docker run -v $(pwd):/home/gradle/`, which needs a Docker daemon and a host-resolvable workspace path. A container job has neither.

### In this PR
* Run the publish job inside `mobilecoin/android-sdk` as the job container, matching the android-bindings publish job.
* Call gradle through a new `publishDirect` makefile target. The docker-wrapped `publish` target calls it too, so publishing keeps one definition.
* Add `build-docker-dispatch.yaml` to publish that builder image, mirroring android-bindings. It rebuilds on a `docker/**` change to master. Only master moves `:latest`.
* Add `permissions`, a per-tag `concurrency` group, `timeout-minutes`, and an `actionlint` config declaring the two self-hosted runner labels.

Notes:
* `make publishDirect` is verified inside the image on `linux/amd64`. `clean` and `assemble` complete and `publish` reaches the Cloudsmith PUT for `com/mobilecoin/android-sdk/6.1.4/android-sdk-6.1.4.aar`, which is what proves `MAVEN_USER` and `MAVEN_PASSWORD` reach the credentials block.
* A dead or empty Cloudsmith credential fails inside `gradle publish` as an HTTP 401, after `clean` and `assemble` have already run, rather than being caught up front. An earlier revision probed the credential before the build; that step was dropped to keep the job close to the android-bindings shape.
* `mobilecoin/android-sdk` exists on Docker Hub and holds no tags. `build-docker-dispatch.yaml` is not dispatchable until it reaches master, and this branch changes nothing under `docker/**`, so the merge does not build the image on its own.
* A cold `linux/amd64` build of `docker/Dockerfile` takes under a minute on a workstation, so `timeout-minutes: 60` sits far above the real cost.
* The publish job on this branch reaches the runner and fails only on the missing image, so the runner label and the container reference are both good.

### Future Work
* Make tests and publish use the same image
* Close the tag-guard gap on `workflow_dispatch`, and run `actionlint` in CI

### DIFF
--- a/.github/workflows/android-sdk-dispatch.yml
+++ b/.github/workflows/android-sdk-dispatch.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: mco-dev-large-x64
     timeout-minutes: 60
     container:
-      image: mobilecoin/android-sdk-builder:latest
+      image: mobilecoin/android-sdk:latest
 
     steps:
       - name: Checkout
--- a/.github/workflows/android-sdk-dispatch.yml
+++ b/.github/workflows/android-sdk-dispatch.yml
@@ -32,43 +32,6 @@ jobs:
             echo "tag ${GITHUB_REF_NAME} would publish version ${version}"
             exit 1
           fi
-      # Proves the credentials are accepted for read. Cloudsmith can scope read
-      # and write separately, so this rules out a dead credential in seconds
-      # without proving the publish itself will be authorised.
-      - name: Check Cloudsmith accepts the credentials
-        shell: bash
-        env:
-          MAVEN_USER: ${{ secrets.MAVEN_USER }}
-          MAVEN_PASSWORD: ${{ secrets.MAVEN_PASSWORD }}
-        run: |
-          # curl reads the credentials from a config on stdin, which keeps them
-          # off the process command line and, unlike a netrc file, does not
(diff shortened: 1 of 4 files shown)


.github/workflows/android-sdk-dispatch.yml
```diff
@@ -6,25 +6,36 @@ on:
       - 'v*'
   workflow_dispatch:
```
stevenspiel, line 7: just noting that we will need to manually run this workflow once merged. `android-sdk-dispatch.yml` uses `mobilecoin/android-sdk:latest`, which won't exist until this is run

makefile
```diff
@@ -62,7 +67,7 @@ publish: setup
 			-e MAVEN_USER \
 			-e MAVEN_PASSWORD \
```
stevenspiel: is the if/else necessary? If these are unset, I don't think Docker passes them at all

stevenspiel (approved):
LGTM

PR #157, opened 2026-08-26
derive TxOut public keys without building a transaction

stevenspiel:
Currently, the only way to get a txout's public key is by calling `prepareTransaction`, which selects inputs and consequently needs a funded, synced account. But it shouldn't need that.

[`TransactionBuilder.addInput`](https://github.com/mobilecoinofficial/android-sdk/blob/18590a27a3baba3c483ccfb20d3ed0f79ab6db11/android-sdk/src/main/java/com/mobilecoin/lib/TransactionBuilder.java#L68) takes no RNG — only [`addOutput`](https://github.com/mobilecoinofficial/android-sdk/blob/18590a27a3baba3c483ccfb20d3ed0f79ab6db11/android-sdk/src/main/java/com/mobilecoin/lib/TransactionBuilder.java#L100) and [`addChangeOutput`](https://github.com/mobilecoinofficial/android-sdk/blob/18590a27a3baba3c483ccfb20d3ed0f79ab6db11/android-sdk/src/main/java/com/mobilecoin/lib/TransactionBuilder.java#L130) receive one. Ring selection consumes the *caller's* stream, separately, at [`MobileCoinClient.java:854`](https://github.com/mobilecoinofficial/android-sdk/blob/18590a27a3baba3c483ccfb20d3ed0f79ab6db11/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java#L854). So the keys produced by an RNG seed do not need (and have never needed) to know what is being spent.

## What this adds

`MobileCoinClient.getTxOutContexts`, which builds against the same `TransactionBuilder`, adds the payload and change outputs, but never calls `build()`. It returns a new `TxOutContexts` object which contains both change and output pub keys. Passing the same RNG to `prepareTransaction` afterwards yields the same two keys.

## Why it's an extraction, not a parallel path

The point is that the two paths *cannot* disagree, so three shared dependencies are now extracted into shared functions (rather than duplicated):

| Extracted | Why it matters to the keys |
| --- | --- |
| `addOutputs` | Holds both output calls and the [block-version change branch](https://github.com/mobilecoinofficial/android-sdk/blob/18590a27a3baba3c483ccfb20d3ed0f79ab6db11/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java#L962), so every draw the builder makes happens in one method |
| `newBuilderSeed` | Holds the hop at [`MobileCoinClient.java:824`](https://github.com/mobilecoinofficial/android-sdk/blob/18590a27a3baba3c483ccfb20d3ed0f79ab6db11/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java#L824) where 32 bytes come off the caller's RNG to seed the builder — the builder never runs on the caller's stream |
| `reportUrisFor` | Which fog reports resolve decides real hint vs fake hint, and those draw different amounts |

That last one is not theoretical. In core, [`create_fog_hint` draws before the tx private key](https://github.com/mobilecoinfoundation/mobilecoin/blob/fe8017c7067b6c4c116f353b9e1beb3db5639888/transaction/builder/src/transaction_builder.rs#L670), and [`r` is drawn immediately after](https://github.com/mobilecoinfoundation/mobilecoin/blob/fe8017c7067b6c4c116f353b9e1beb3db5639888/transaction/builder/src/transaction_builder.rs#L681) — so a resolver built differently shifts every key that follows.

## The zero change amount

The derivation reports a change amount of zero, having no inputs to leave a remainder. That is only sound because nothing between the builder's two draws consumes randomness, so neither amount can move a key.

`testAmountsDoNotChangeTheKeys` tests that one seed and two very different amounts both return the same keys.

## Requirements

No balance, but the user must be online. Fog reports and the network block version are fetched, and a synced `TxOutStore` supplies the block index those reports are validated against. It can fail like any other fog operation, so callers need a retry path rather than treating it as local computation.

## Also here: the funding diagnostic fires on a spent key image

Unrelated to the derivation, carried here because this branch is where it surfaced.

`test_fragmented_account` fails on this branch with `InvalidTransactionException: ContainsSpentKeyImage`, and #158's diagnostic stayed silent because it only matched `InsufficientFundsException`. Both facts have the same cause. `TestKeysManager` hands out accounts by a rotating index, this branch adds seven tests that each build a client — 211 tests to 218 — so every draw after them shifted and the test drew a different, starved account than the one funded for #158. A wallet down to a single output has it spent by the first transaction in the loop, then prepares the next against a view that has not yet seen the change and re-selects the output it just spent. Consensus rejects it for the spent key image, so a funding problem arrives wearing the name of a correctness bug.

[`looksLikeStarvedWallet`](https://github.com/mobilecoinofficial/android-sdk/blob/e76e9a3/android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java#L167) now matches that too, against [the specific `ProposeTxResult`](https://github.com/mobilecoinofficial/android-sdk/blob/e76e9a3/android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java#L174) rather than every invalid transaction, so a tombstone-expired or fee-too-low rejection still passes through untouched. The balances already in the report are what separate the two: healthy balances under a spent key image mean it really is a double spend.

[README § Test wallets](https://github.com/mobilecoinofficial/android-sdk/blob/e76e9a3/README.md#L102) records what that section left out — the suite spends real balance from a fixed, unreplenished set of accounts, and because the index rotates, the account a test draws changes whenever tests are added. Funding whichever wallet failed last time is not a durable fix; keeping every configured account funded is.


android-sdk/src/main/java/com/mobilecoin/lib/TxOutContexts.java
```diff
@@ -0,0 +1,39 @@
...
+import androidx.annotation.NonNull;
+
+/**
+ * The payload and change outputs a transaction's RNG seed produces.
+ * <p>
+ * Returned on its own by
+ * {@link MobileCoinClient#getTxOutContexts(PublicAddress, Amount, Amount, TxOutMemoBuilder, Rng)},
+ * which derives the outputs without selecting inputs or building a
+ * transaction, and carried inside {@link PendingTransaction} when a
+ * transaction is actually built.
```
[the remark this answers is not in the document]
stevenspiel: that's fair. updated in 3b8ee1d46d0bee4ab5bdca1eb4030d8969ac6d58

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java
```diff
@@ -0,0 +1,178 @@
...
+        final Amount amountToSend = Amount.ofMOB(BigInteger.valueOf(52398457942L));
+        final Amount fee = client.estimateTotalFee(amountToSend);
+        final PublicAddress recipient = TestKeysManager.getNextAccountKey().getPublicAddress();
+
+        final TxOutContexts first = client.getTxOutContexts(recipient, amountToSend, fee,
+                TxOutMemoBuilder.createSenderAndDestinationRTHMemoBuilder(client.getAccountKey()),
+                ChaCha20Rng.fromSeed(SEED));
+        final TxOutContexts second = client.getTxOutContexts(recipient, amountToSend, fee,
+                TxOutMemoBuilder.createSenderAndDestinationRTHMemoBuilder(client.getAccountKey()),
+                ChaCha20Rng.withRandomSeed());
```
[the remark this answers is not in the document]
stevenspiel: done 3b8ee1d46d0bee4ab5bdca1eb4030d8969ac6d58

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -804,6 +804,158 @@ public PendingTransaction prepareTransaction(
...
+            if (recipient.hasFogInfo()) {
+                reportUris.add(new FogUri(recipient.getFogReportUri()));
+            }
+            reportUris.add(new FogUri(getAccountKey().getFogReportUri()));
+        } catch (InvalidUriException exception) {
+            FogReportException reportException = new FogReportException("Invalid Fog Report " +
+                    "Uri in the public address");
+            Util.logException(TAG, reportException);
+            throw reportException;
+        }
```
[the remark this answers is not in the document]
stevenspiel: done 3b8ee1d46d0bee4ab5bdca1eb4030d8969ac6d58

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java
```diff
@@ -0,0 +1,155 @@
...
+                built.getChangeTxOutContext().getTxOutPublicKey(),
+                derived.getChange().getTxOutPublicKey());
+    }
+
+    @Test
+    public void testSameSeedDerivesTheSameKeys() throws Exception {
+        final ChaCha20Rng rng = ChaCha20Rng.fromSeed(SEED);
+        final MobileCoinClient client = MobileCoinClientBuilder.newBuilder().build();
+        final Amount amountToSend = Amount.ofMOB(BigInteger.valueOf(52398457942L));
+        final Amount fee = client.estimateTotalFee(amountToSend);
```
bcgreijnlautier: suggestion: `amountToSend` and `fee` are dead now that `getTxOutContexts` takes only the recipient and the rng, and `estimateTotalFee` is a network round trip for a value nothing reads.

```suggestion
```

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java
```diff
@@ -0,0 +1,155 @@
...
+        assertEquals(
+                first.getChange().getTxOutPublicKey(),
+                second.getChange().getTxOutPublicKey());
+    }
+
+    @Test
+    public void testDifferentSeedsDeriveDifferentKeys() throws Exception {
+        final MobileCoinClient client = MobileCoinClientBuilder.newBuilder().build();
+        final Amount amountToSend = Amount.ofMOB(BigInteger.valueOf(52398457942L));
+        final Amount fee = client.estimateTotalFee(amountToSend);
```
bcgreijnlautier: Same here.

```suggestion
```

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java
```diff
@@ -0,0 +1,155 @@
...
+        final PublicAddress recipient = TestKeysManager.getNextAccountKey().getPublicAddress();
+
+        final TxOutContexts first = client.getTxOutContexts(recipient, ChaCha20Rng.fromSeed(SEED));
+        final TxOutContexts second = client.getTxOutContexts(recipient, ChaCha20Rng.fromSeed(OTHER_SEED));
+
+        assertNotEquals(
```
bcgreijnlautier, line 147: suggestion: the Swift equivalent of this test asserts the change keys differ as well, and the change output draws off the same stream. I'd assert both here so the two suites stay symmetric.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * The public key is {@code r * D}, so the recipient is as much an input to
+     * it as the seed. A caller holding only a seed cannot know the key.
+     */
+    @Test
+    public void testRecipientChangesThePayloadKey() throws Exception {
+        final ChaCha20Rng rng = ChaCha20Rng.fromSeed(SEED);
+        final MobileCoinClient client = MobileCoinClientBuilder.newBuilder().build();
+        final Amount amountToSend = Amount.ofMOB(BigInteger.valueOf(52398457942L));
+        final Amount fee = client.estimateTotalFee(amountToSend);
```
bcgreijnlautier: Here as well.

```suggestion
```

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -804,6 +804,160 @@ public PendingTransaction prepareTransaction(
...
+        return new TxOutContexts(payloadTxOutContext, changeTxOutContext);
+    }
+
+    /**
+     * Derives the payload and change outputs a transaction from {@code rng}
+     * would produce, without selecting inputs or building one.
+     *
+     * <p>Intended for learning a TxOut public key before the transaction
+     * exists — for instance to tell a third party which output to expect.
+     * Passing the same {@code rng} seed to
```
bcgreijnlautier: suggestion: `newBuilderSeed` has already taken 32 bytes off `rng` by the time this returns, which is why every test here rewinds with `rng.setWordPos(BigInteger.ZERO)` before the second call. `Rng` exposes no rewind of its own, so a caller who hands the same instance straight to `prepareTransaction` gets different keys and would tell a third party the wrong TxOut. I'd state here that the rng has to be back at the same word position, or rebuilt from the seed.

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
```diff
@@ -804,6 +804,160 @@ public PendingTransaction prepareTransaction(
...
+        // Fetched once and reused below: a refresh between constructing the
+        // builder and choosing the change output would build for one block
+        // version and branch on another.
+        final int blockVersion = blockchainClient.getOrFetchNetworkBlockVersion();
+
+        // No amount reaches a draw, so zero gives the same keys as any real
+        // transaction. Both outputs share a token id so the builder's
+        // mixed-token check cannot trip on values that describe no real
+        // transaction anyway.
+        final Amount zero = new Amount(BigInteger.ZERO, TokenId.MOB);
```
bcgreijnlautier: suggestion: this pins every derivation to `TokenId.MOB`, and so does the token id passed to the `TransactionBuilder` constructor below, so deriving for a non-MOB transaction builds against a token id the real send will not use. Nothing in this suite varies the token id, so its irrelevance to the keys is assumed rather than pinned the way the amount and the memo are. I'd either take the token id as a parameter or add a test that varies it.

android-sdk/src/main/java/com/mobilecoin/lib/TxOutContexts.java
```diff
@@ -0,0 +1,39 @@
+package com.mobilecoin.lib;
+
+import androidx.annotation.NonNull;
+
+/**
+ * The payload and change outputs a transaction's RNG seed produces.
+ * <p>
+ * Returned by
+ * {@link MobileCoinClient#getTxOutContexts(PublicAddress, Amount, Amount, TxOutMemoBuilder, Rng)},
```
bcgreijnlautier: Same here.

```suggestion
 * {@link MobileCoinClient#getTxOutContexts(PublicAddress, Rng)},
```

bcgreijnlautier (approved):

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
stevenspiel: yep. thanks for catching that on the iOS side, and it's bee resolved on this side, too. 

The tests called `rng.setWordPos(BigInteger.ZERO)` between the two calls, so this wasn't getting caught. I fixed it in 8237976.

android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
stevenspiel: fixed in 096986a

android-sdk/src/main/java/com/mobilecoin/lib/TxOutContexts.java
stevenspiel: yeah, the signature has changed since. updated in 8237976

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java
stevenspiel: done in f764066

PR #158, opened 2026-08-26
name the wallet to fund when a test runs out of funds

stevenspiel:
## Problem

Two things were invisible.

**CI printed no failure messages at all.** `gcloud firebase test android run` streams a summary table and nothing else, so [run 33025432166](https://github.com/mobilecoinofficial/android-sdk/actions/runs/33025432166) reported the entire outcome as `3 test cases failed, 208 passed`. Every message sat in the GCS bucket gcloud names on the way in, which nobody opens.

**Nothing said which wallet to top up.** `MobileCoinClientTest.test_fragmented_account`, `MobileCoinClientTest.test_post_to_serialized_public_address` and `TransactionBuilderTest.testPredictableOutputPublicKey` have been red with `InsufficientFundsException` since Aug 25, on master as well as feature branches. The suite spends real balance on every run, and [`TestKeysManager`](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/android-sdk/src/androidTest/java/com/mobilecoin/lib/TestKeysManager.java#L85) hands out accounts with a rotating index, so which one a test draws depends on how many built a client before it and shifts whenever tests are added or reordered. GitHub secrets are write-only, so the addresses only exist inside a run. `InsufficientFundsException` carries no state either — not the wallet, not the token.

## What this adds

[`report_results`](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/docker/scripts/run_connected_tests.sh#L15) captures the bucket path gcloud prints, pulls the JUnit XML back, and prints it to the CI console, [painting failures red](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/docker/scripts/run_connected_tests.sh#L34) since the XML is mostly passing testcases. This is not funding-specific: every failure message is now readable in the log.

[`FundingDiagnosticRule`](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java#L42) checks a failure's cause chain for `InsufficientFundsException` and, only then, [reports the balances](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java#L60) of the accounts that test drew. It [rewraps the failure](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java#L53) with the report appended, keeping the original as the cause so its type and stack survive. `TestKeysManager` [records what it issues](https://github.com/mobilecoinofficial/android-sdk/blob/name-the-wallet-to-fund/android-sdk/src/androidTest/java/com/mobilecoin/lib/TestKeysManager.java#L29) so the rule knows which accounts to ask about.

Failing tests include output like:

```
<failure>java.lang.AssertionError: Insufficient funds
Out of funds on TEST_NET. This test drew:
  2hWLwBrWAnoFAkatWtiXNbtmuZUNNrcQiAsWGfetuV5i2PfQ441LPZ7w2NuVfMURvTSQ1ubcb2YivAhGFUZJSm9HHp7rbJJA1EqHXRZ2ikmh129qaP6aDVqGubaYRMTgTgoQzRChQ4632tNGGnYNngT1Bv9QUJqhAaG8f84EhCFdxZVL73TqkvcfKaPYvAMJY9D3RqNyRh11T33PAFo8Vi1yRnX8AGfcJxm7D1T2H1eMdhSyQ
    holds 0.000000002982 MOB
  t5Tx8WUudopcPAhaTRFcX3UL1wmA3FQtSYJt1cyj7GTYErsN5TRMQKBiTp8CnBsRRVNFBQBdLNWbN8KLkUpzwA6vPoRVD5EDQsfiZcqDHVKeauDbm84K5VUnXrNtEzeCFQxQgP7PJkFbijv9RsTNUW2y8YdALRZi9fTgim6ShbXoN3dWzTzTvKTVmDonuvqG9yb4LqZRno9HqtSuTMzirRxWwHQSTS88BwRDRkYTGK5mYZkk
    holds 62.235157890776 MOB
  2o1EFeXi6LtBXBUJ5rvocUwN3TbZLW7Sdz1uzTWfUZ8A3TWUkRKFk8hrfkZytkt4pxe3XfGaHkAjcrhySZykFGkH9n9voRkHyJKcJiEhSJXNijiznMsA8RrxLjWQXKiQN11Z9ij6uM19HCxYYZ54oi2VaaVuu4WFiB4cNeDDnzLGcxZg4RdiVEyysE53AAESayWVko8misMkGfFYfnQi2rcaBLqzdmsie7hqG9try2gk3eUS4
    holds 0.97640000262 MOB
```

## Cost

Nothing on a green run. The balances are read inside the failure, on a client built for the occasion, for the one to three accounts that test drew — verified as the maximum any single test draws.

The alternative — sweeping every wallet up front — was measured at 205s on a ~16 minute run, on every run including green ones. That version is what produced the address above: exactly one wallet below 0.1 MOB, and it accounts for all three failures, because the index wraps every ten draws so the same drained account keeps coming back around.

Running the check only after a failure in a second Test Lab matrix was tried and does not work: a second matrix is a second test execution against a daily quota, and it fails validation with `TEST_QUOTA_EXCEEDED` precisely on the days something is failing.

## Notes

**No token is guessed at.** Every failing call is `Amount.ofMOB`, and an earlier version that accepted a non-zero balance in any token passed while every spending test failed — the wallets hold other tokens. Reporting balances per token rather than asserting a threshold avoids encoding what each test needs.

**Addresses only, never mnemonics.** This repo is public, so Actions logs are world-readable. A b58 address is what you hand someone to receive funds; a mnemonic would be a published private key.

**This reports the symptom.** The cause is that the whole instrumented suite, transacting tests included, runs on every push to every branch, against ten fixed wallets nothing replenishes. Gating the spending tests to manual dispatch would fix the drain but lose per-push coverage of the spending paths, so it is deliberately not proposed here.

**The rule is a `@Rule`, declared once on a base class.** JUnit's only run-wide hook is RunListener, which can observe a failure but not modify it, and observation lands in logcat rather than the JUnit XML — so a `@Rule` is what puts the report where CI can read it. JUnit picks up `@Rule` fields from a superclass, so the twelve spending test classes extend FundedTest rather than each declaring their own.

**Only MOB is converted to whole coin.** Decimals are per-token and the SDK carries none of them — Amount and TokenId have no scale field. MOB is twelve, eUSD is six, so every non-MOB balance is quoted in its smallest unit rather than guessed at.


stevenspiel:
Once the PR is approved, I'll fund the wallet so these tests can pass again.

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
```diff
@@ -0,0 +1,149 @@
...
+            // Diagnostics must not replace the failure that got us here.
+            return "<balance unavailable: " + exception + ">";
+        } finally {
+            if (client != null) {
+                client.shutdown();
+            }
+        }
+    }
+
+    /** Every token the suite spends is quoted in the same twelve decimals. */
```
bcgreijnlautier: issue: `balancesOf` reports every token the account holds, not only the ones the suite spends, and `asAmount` divides all of them by MOB's twelve decimals. eUSD is six-decimal, `SignedContingentInputTest` quotes 10 eUSD as `10000000`, so a wallet holding 10 eUSD prints `0.00001 eUSD` and reads as drained. I'd convert only for `TokenId.MOB` and print the raw value with the token name for anything else.

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
```diff
@@ -0,0 +1,149 @@
...
+    private static String asAmount(final BigInteger value) {
+        if (value.signum() == 0) {
+            return "0";
+        }
```
bcgreijnlautier: nit: `balancesOf` skips a zero balance with `continue` before this is reached, so the branch is unreachable.

```suggestion
```

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
```diff
@@ -0,0 +1,149 @@
...
+        if (issued.isEmpty()) {
+            return message.append(" No account was issued to this test.").toString();
+        }
+
+        // Every account the test drew, balance and all. The one to fund is the
+        // one sitting at nothing, so there is no need to guess at how each was
+        // used — a recipient that never spends reads as funded either way.
+        message.append(" This test drew:");
+        for (final AccountKey accountKey : issued.values()) {
+            message.append("\n  ").append(addressOf(accountKey))
```
bcgreijnlautier: suggestion: `issuedKeys` is keyed by the account's index in the configured list and the report drops it, so whoever funds the wallet has to match a b58 address back to a mnemonic by hand. I'd iterate `entrySet()` and print the index next to the address.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TestKeysManager.java
```diff
@@ -2,18 +2,47 @@
...
+     */
+    private static final Map<Integer, AccountKey> issuedKeys = new LinkedHashMap<>();
+
+    static void forgetIssuedKeys() {
+        issuedKeys.clear();
+    }
+
+    @NonNull
+    static Map<Integer, AccountKey> getIssuedKeys() {
+        return Collections.unmodifiableMap(issuedKeys);
```
bcgreijnlautier: suggestion: this hands back a live view of the mutable static map, and `walletsToFund` iterates it while building a client per account. A draw that lands during the report would turn it into a `ConcurrentModificationException` that replaces the failure being diagnosed, so I think it's worth returning a copy instead.

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountSnapshotTest.java
```diff
@@ -9,12 +9,17 @@
...
+import org.junit.Rule;
 import org.junit.Test;
 
 import java.math.BigInteger;
 
 public class AccountSnapshotTest {
 
+    /** Names the wallet to fund when a test fails for want of funds. */
+    @Rule
+    public final FundingDiagnosticRule fundingDiagnostic = new FundingDiagnosticRule();
```
bcgreijnlautier: suggestion (optional): I'd rather this lived on one abstract test class these twelve extend, since JUnit picks up `@Rule` fields declared on a superclass and that drops eleven copies of the field and its comment. Ultimately your call though.

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
stevenspiel: good thinking. done here e37f11fc7f3c5adb4ada4bb620e0849a23c62699

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
stevenspiel: e8811591a579116de3f2b659717f5658fbcd40cd

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
stevenspiel: Index now getting printed. 98c083c47dd2bd56a835540f58834cb73bbb6f38

android-sdk/src/androidTest/java/com/mobilecoin/lib/TestKeysManager.java
stevenspiel: yeah, good catch 58fbc35a381a2ed360f22219650adabda88ed9f2

android-sdk/src/androidTest/java/com/mobilecoin/lib/AccountSnapshotTest.java
stevenspiel: I like that better d4e4a5460afa4b588ec0b2261481b59b50722526

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
```diff
@@ -0,0 +1,149 @@
...
+            public void evaluate() throws Throwable {
+                TestKeysManager.forgetIssuedKeys();
+                try {
+                    base.evaluate();
+                } catch (final Throwable failure) {
+                    if (!causedByInsufficientFunds(failure)) {
+                        throw failure;
+                    }
+                    throw new AssertionError(
+                            failure.getMessage() + "\n\n" + walletsToFund(), failure);
```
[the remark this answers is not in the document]
stevenspiel: 64cc30102db04097bd6dbc8602d0c71740f2bdfe

android-sdk/src/androidTest/java/com/mobilecoin/lib/FundingDiagnosticRule.java
```diff
@@ -0,0 +1,149 @@
...
+ * the configured accounts, so which one a given test drew is not recoverable
+ * from the failure alone — and the answer shifts as tests are added or
+ * reordered. {@link InsufficientFundsException} carries no state either, so
+ * the token is not recoverable, which is why this reports balances rather
+ * than only addresses.
+ * <p>
+ * It reads those balances inside the failure, on a client the test has
+ * already built and synced, so a passing run pays nothing at all. Sweeping
+ * every wallet up front costs about 20s each on every run, green ones
+ * included; this costs that only for the accounts a failing test drew.
```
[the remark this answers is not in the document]
stevenspiel: f827808d2e211c1026236f7a3b063418f7e74ad8

stevenspiel:
Output is now

```
<failure>java.lang.AssertionError: Insufficient funds

Out of funds on TEST_NET. This test drew:
  account 3: jCDqZsEboYy8qhpJ9QKVp6ioVfm3quc6LFcgk9A4n3inawAuu4YoXE8TJDG3KBkwahPLFuj5SS3z1YTNspY8xVeHSSd9o24tPrCh8tCa8LQZwoWSLWBqER43fK5ST44qE8yscjUAdT5wThe225KfHHyDxji22Az8bAjEDybKpfTnWT1gKa47EzDRzjuwVHRymYVq3vsa2fhX4AmvV2hZ6Rmg4y6qJpBup7zrCxM4GJgC8S5q
    holds 0.944799918105 MOB
  account 4: 2hWLwBrWAnoFAkatWtiXNbtmuZUNNrcQiAsWGfetuV5i2PfQ441LPZ7w2NuVfMURvTSQ1ubcb2YivAhGFUZJSm9HHp7rbJJA1EqHXRZ2ikmh129qaP6aDVqGubaYRMTgTgoQzRChQ4632tNGGnYNngT1Bv9QUJqhAaG8f84EhCFdxZVL73TqkvcfKaPYvAMJY9D3RqNyRh11T33PAFo8Vi1yRnX8AGfcJxm7D1T2H1eMdhSyQ
    holds 0.000000003056 MOB
```

bcgreijnlautier (approved):

stevenspiel:
sent 0.007 testnet MOB to `2hWLwBrWAnoFAkatWtiXNbtmuZUNNrcQiAsWGfetuV5i2PfQ441LPZ7w2NuVfMURvTSQ1ubcb2YivAhGFUZJSm9HHp7rbJJA1EqHXRZ2ikmh129qaP6aDVqGubaYRMTgTgoQzRChQ4632tNGGnYNngT1Bv9QUJqhAaG8f84EhCFdxZVL73TqkvcfKaPYvAMJY9D3RqNyRh11T33PAFo8Vi1yRnX8AGfcJxm7D1T2H1eMdhSyQ` and triggered CI

stevenspiel:
CI failed, but not with an InsufficientFunds exception, which is progress. Investigating...

PR #159, opened 2026-08-31
fix: scope HTTP Authorization header to configured hosts

stevenspiel:
## Problem

Fog report URLs are read from the recipient's public address, so the host is chosen by the counterparty:

- `FogReportsManager.java#L97-L99` builds a `ReportClient` per counterparty URI
- `RestServiceAPIManager.java#L44-L46` hands that URI the caller's shared `Requester`

So a consumer on `TransportProtocol.forHTTP(new HttpRequester(user, pass))` sends their credentials to whatever host is in an address someone sends them. That wiring is a documented option and is what `MobileCoinClientWithHttpRequesterTest` does.

## Fix

Scope the header to the hosts the client was configured with. `MobileCoinClient` already has the normalized fog and consensus URIs in hand, so it passes them to the requester at construction:

```java
scopeHttpAuthorization(transportProtocol, normalizedFogUri, normalizedConsensusUris);
```

**No consumer code change** — the allowlist is derived from URIs callers already pass in, so existing integrations are fixed on upgrade.

Hosts accumulate rather than replace, so one requester shared across several clients keeps working.

### DIFF
--- a/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
+++ b/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java
@@ -28,6 +28,8 @@ import com.mobilecoin.lib.log.LogAdapter;
 import com.mobilecoin.lib.log.Logger;
 import com.mobilecoin.lib.network.NetworkResult;
 import com.mobilecoin.lib.network.TransportProtocol;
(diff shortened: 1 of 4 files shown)


stevenspiel:
Closing this as a whitelist is not the right approach here.

PR #160, opened 2026-09-08
test: assert the seed vector MobileCoin-Swift derives

stevenspiel:
### Motivation

Each SDK currently proves only that it agrees with itself. [`TxOutContextsTest`](https://github.com/mobilecoinofficial/android-sdk/blob/fbd52bc1541379a2c1fa30f172cbc0d59b60772f/android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsTest.java) pins that [`getTxOutContexts`](https://github.com/mobilecoinofficial/android-sdk/blob/fbd52bc1541379a2c1fa30f172cbc0d59b60772f/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java#L930) ignores amounts, fees, memos, token id and block version, and that one seed run twice gives one answer. MobileCoin-Swift's [`TxOutContextsDerivationTests`](https://github.com/mobilecoinofficial/MobileCoin-Swift/blob/cc5ebc9a0d273144ba9efbe9d05484e94643974a/Tests/Unit/Transaction/TxOutContextsDerivationTests.swift) pins the same properties over there. Both suites would keep passing if the two SDKs derived different keys from the same seed: they assert relationships between keys rather than any particular key, and this one uses `TestKeysManager.getNextAccountKey()`, which rotates, so there has never been anything for iOS to match.

### In this PR
* Adds `TxOutContextsParityTest`, asserting the payload and change TxOut public keys a fixed seed derives for a fixed sender and recipient, against [the constants MobileCoin-Swift asserts for the same inputs](https://github.com/mobilecoinofficial/MobileCoin-Swift/commit/db2f7488b5817e4f12b03e42bb5e2386a805aeb3).
* The sender and recipient are deserialized from serialized `AccountKey` and `PublicAddress` protobufs — the same bytes the Swift test reads — rather than from a rotating test account, so the two sides demonstrably run the same inputs.
* Uses the seed already in `TxOutContextsTest`, so a divergence between the two tests is a divergence in the derivation rather than in the input.
* Calls the [package-private `getTxOutContexts` overload](https://github.com/mobilecoinofficial/android-sdk/blob/fbd52bc1541379a2c1fa30f172cbc0d59b60772f/android-sdk/src/main/java/com/mobilecoin/lib/MobileCoinClient.java#L957) to pin block version and token id rather than read them from the network, so the vector does not move when the network does. Both are already covered as irrelevant to the keys.

This goes with the [MobileCoin-Swift PR](https://github.com/mobilecoinofficial/MobileCoin-Swift/pull/321). Neither half is meaningful alone.


android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Block version and token id are pinned rather than read from the network,
+     * so the vector does not move when the network does. Both are covered as
+     * irrelevant to the keys by {@link TxOutContextsTest}; pinning them here
+     * only keeps this test asserting one thing.
+     */
+    @Test
+    public void testSeedDerivesTheKeysSwiftDerives() throws Exception {
+        final AccountKey accountKey =
+                AccountKey.fromBytes(Base64.decode(ACCOUNT_KEY_B64, Base64.DEFAULT));
```
bcgreijnlautier: question: this account key and the recipient both carry `fog://fog.test.mobilecoin.com`, while the client below takes its fog uri and trusted identities from `Environment.CURRENT_TEST_ENV`. That constant is `TEST_NET` today, so flipping it to `MOBILE_DEV` to run the suite against devnet would fail this test on attestation, while every test drawing from `TestKeysManager` keeps following the env. Is an `assumeTrue` on the env worth adding here, so the failure names the reason?

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
```diff
@@ -0,0 +1,108 @@
...
+        final PublicAddress recipient =
+                PublicAddress.fromBytes(Base64.decode(RECIPIENT_B64, Base64.DEFAULT));
+
+        final MobileCoinClient client =
+                MobileCoinClientBuilder.newBuilder().setAccountKey(accountKey).build();
```
bcgreijnlautier: nit: `newBuilder()` draws `TestKeysManager.getNextAccountKey()` in its constructor before `setAccountKey` replaces it, so this advances the rotation for every later test in the run and records an account in `issuedKeys` that this test never touches. `FundingDiagnosticRule` reads that map, so an unrelated funding failure would report it.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
```diff
@@ -0,0 +1,123 @@
...
+ * edited on one side alone, the vector proves nothing.
+ *
+ * <p>Two inputs are not shared, and neither reaches a draw. The two sides
+ * reach their fog reports differently — Swift resolves a canned report
+ * offline, this fetches TestNet live — because only Swift has a seam below fog
+ * resolution. That is deliberate and is not a gap: the report's contents
+ * change which key the hint is encrypted to, never how many draws encrypting
+ * it takes, so the draw that produces {@code r} lands in the same place either
+ * way. The tombstone block index differs for the same reason — this derives it
+ * from the live chain tip, Swift takes a fixture's — and it reaches a key even
```
bcgreijnlautier: suggestion: I don't think this reads the live chain tip. `getTxOutContexts` derives the tombstone from `txOutStore.getCurrentBlockIndex()`, which stays `UnsignedLong.ZERO` until `refresh` runs, and `newBuilder()` supplies no `StorageAdapter` so no cached store is restored either. That leaves it at a constant `DEFAULT_NEW_TX_BLOCK_ATTEMPTS` of 50, so I'd describe it as a pinned input here.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
```diff
@@ -0,0 +1,123 @@
...
+            "xArSo1TSmrdSOO5z/ChxT1A/asuGT7z3/4I0tTylKH4=";
+    private static final String EXPECTED_CHANGE_KEY_B64 =
+            "vnTuxXXogiaGAWcaU5HUZ1VYzaP+mjnJbMudemo3qAA=";
+
+    /**
+     * Block version and token id are pinned rather than read from the network,
+     * so neither can move the vector when the network moves. Both are covered
+     * as irrelevant to the keys by {@link TxOutContextsTest}; pinning them here
+     * only keeps this test asserting one thing. The tombstone block index is
+     * still read from the live chain — there is no entry point that takes it —
```
bcgreijnlautier: Same here.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
```diff
@@ -0,0 +1,123 @@
...
+@RunWith(AndroidJUnit4.class)
+public class TxOutContextsParityTest {
+
+    /**
+     * Arbitrary, and fixed only so both platforms draw from the same stream.
+     * 32 bytes: 31 of ASCII and a trailing zero. The same bytes as
+     * {@code TxOutContextsTest}'s seed, so a divergence between the two tests
+     * is a divergence in the derivation rather than in the input.
+     */
+    private static final byte[] SEED = new byte[] {103, 111, 116, 111, 32, 104, 116, 116, 112,
```
bcgreijnlautier: suggestion: nothing enforces that these bytes stay equal to `TxOutContextsTest`'s seed, so an edit to one file will silently break the invariant the comment above states. Both classes sit in `com.mobilecoin.lib`, so I'd drop `private` on `TxOutContextsTest.SEED` and reference it here.

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    @Test
+    public void testSeedDerivesTheKeysSwiftDerives() throws Exception {
+        final AccountKey accountKey =
+                AccountKey.fromBytes(Base64.decode(ACCOUNT_KEY_B64, Base64.DEFAULT));
+        final PublicAddress recipient =
+                PublicAddress.fromBytes(Base64.decode(RECIPIENT_B64, Base64.DEFAULT));
+
+        final MobileCoinClient client =
+                MobileCoinClientBuilder.newBuilder().setAccountKey(accountKey).build();
```
bcgreijnlautier: suggestion: `newBuilder()` draws a rotating account key through `TestKeysManager.getNextAccountKey()` in its private constructor, so `setAccountKey` discards an account the rotation has already handed out and recorded. `FundingDiagnosticRule` says shifting which account a test draws has caught this suite out twice, so I'd add a `newBuilder` overload taking an `AccountKey` that skips the draw.

bcgreijnlautier (approved):

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
stevenspiel: great idea. done here 9f5eccea92d09655a77e8701d2f40d238c5fe091 and updated 5 other places that needed it 1db8a1eadfb07efb5a0e75415bfbf4dd2c0161c1

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
stevenspiel: 3b9f469fa20117ff95bc98680826893c95c79100

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
stevenspiel: a7721380cd4d84305083a1bb739ed71a7c90fbfb

android-sdk/src/androidTest/java/com/mobilecoin/lib/TxOutContextsParityTest.java
stevenspiel: a7721380cd4d84305083a1bb739ed71a7c90fbfb

